Issue #22, opened 2022-08-09
Undocumented duplicate key behaviour for events and rules


baldawar:
## Describe the bug

High level description of the issue is we can create duplicate keys in our rules and the library will only use the last duplicate in the rule

For example if this is my rule:
```
{
  "source": ["aws.s3"],
  "source": ["aws.sns"],
  "detail-type": ["AWS API Call via CloudTrail"],
  "detail": {
    "eventSource": ["s3.amazonaws.com"],
    "eventSource": ["sns.amazonaws.com"]
  }
}
```

A customer may expect it to capture both detail S3 and SNS events from CloudTrail. But it turns out we only trigger only the second key (which in this case is SNS).

This is within the bounds of the JSON spec https://datatracker.ietf.org/doc/html/rfc8259#section-4 
> When the names within an object are not
unique, the behavior of software that receives such an object is
unpredictable.  Many implementations report the last name/value pair
only.

Still, it's better if we 1/ handle it for users when de-serializing via ObjectMapper[1] and 2/ document this edge case if they are passing the JSONNode to us. 

[1] https://github.com/FasterXML/jackson-databind/issues/237 shows that we can use `DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY`

nicdard:
I have been taking a look at this issue, and tried to use the configuration for the `JSON_FACTORY` in `JsonRuleCompiler` to disable duplicate fields in rules (`[STRICT_DUPLICATE_DETECTION](https://fasterxml.github.io/jackson-core/javadoc/2.9/com/fasterxml/jackson/core/JsonParser.Feature.html#STRICT_DUPLICATE_DETECTION).`).

However this configuration is too strict and disallow valid rules, for example looking at this test:

```java
@Test
    public void intensitiesACTest() throws Exception {
        String intensities = "{\n" +
                "  \"lines\": [{ \n" +
                "    \"intensities\": [\n" +
                "      [\n" +
                "        0.5\n" +
                "      ]\n" +
                "    ],\n" +
                "    \"points\": [\n" +
                "      [\n" +
                "        [\n" +
                "          {\"pp\" : \"index0\"},\n" +
                "          483.3474497412421,\n" +
                "          287.48116291799363\n" +
                "        ],\n" +
                "        [\n" +
                "          {\"pp\" : \"index1\"},\n" +
                "          489.9999497412421,\n" +
                "          299.99996291799363\n" +
                "        ]\n" +
                "        \n" +
[...]

PR #34, opened 2022-09-06
Updating documentation to explain duplicate key deserialization behaviour

baldawar:
### Description of changes:
This minor change is related to
https://github.com/aws/event-ruler/issues/22.
Would like to make sure there's at least something in the document
on the expected behaviour.

In future, this section can be updated with recommendation to
move to better APIs.

Also updating the pom.xml to make it easier to release changes. Going
forward, we'll enforce version bumps for key changes.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -682,8 +682,31 @@ This behaviour may change in future version (to avoid any confusions) and should
 #### Caveat: JSON parsing and duplicate keys
 
 Events or Rules containing duplicate JSON keys are invalid. When duplicate keys are passed Ruler 
-will only consider the final value in the document order. This means following two rules
-will compile to the same internal representation.
+will only consider the final value in the document order. 
+
+This means following two rules will compile to the same internal representation.
+
+```javascript
+## has duplicate keys
+{
+  "source": ["aws.s3"],
+  "source": ["aws.sns"],
+  "detail-type": ["AWS API Call via CloudTrail"],
+  "detail":  {
+      "eventSource": ["s3.amazonaws.com"],
+      "eventSource": ["sns.amazonaws.com"]
+  }
+}
+
+## has unique keys
+{
+  "source": ["aws.sns"],
+  "detail-type": ["AWS API Call via CloudTrail"],
+  "detail": { "eventSource": ["sns.amazonaws.com"] }
+}
+```
+
+Similarly, following two *events* will be treated the same
 
 ```javascript
 ## has duplicate keys
--- a/README.md
+++ b/README.md
@@ -679,11 +679,11 @@ It also means that these rules will match against following two events :
 
 This behaviour may change in future version (to avoid any confusions) and should not be relied upon.
 
-#### Caveat: JSON parsing and duplicate Keys
+#### Caveat: JSON parsing and duplicate keys
 
-When Ruler deserializes Events or Rules containing duplicate JSON keys, it will only consider the
-final value in the document order. This means following two rules will compile to the same
-internal representation.
+Events or Rules containing duplicate JSON keys as invalid. When duplicate keys are passed Ruler 
+will only consider the final value in the document order. This means following two rules
+will compile to the same internal representation.
 
 ```javascript
 ## has duplicate keys
--- a/README.md
+++ b/README.md
@@ -691,9 +691,11 @@ will compile to the same internal representation.
   "source": ["aws.s3"],
   "source": ["aws.sns"],
   "detail-type": ["AWS API Call via CloudTrail"],
-  "detail": [
-    { "eventSource": ["s3.amazonaws.com"] },
-    { "eventSource": ["sns.amazonaws.com"] }
+  "detail":  [
+    {
+      "eventSource": ["s3.amazonaws.com"],
+      "eventSource": ["sns.amazonaws.com"]
+    }
   ]
 }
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -679,6 +679,61 @@ It also means that these rules will match against following two events :
 
 This behaviour may change in future version (to avoid any confusions) and should not be relied upon.
 
+#### Caveat: JSON parsing and duplicate keys
+
+Events or Rules containing duplicate JSON keys are invalid. When duplicate keys are passed Ruler 
+will only consider the final value in the document order. 
+
+This means following two rules will compile to the same internal representation.
```
timbray, line 687: Minor point: I'd say "…will match the same events" because that's what users care about, not the internal representation. Not a blocker.

README.md
```diff
@@ -679,6 +679,61 @@ It also means that these rules will match against following two events :
...
+
+## has unique keys
+{
+  "source": ["aws.sns"],
+  "detail-type": ["AWS API Call via CloudTrail"],
+  "detail": { "eventSource": ["sns.amazonaws.com"] }
+}
+```
+
+Similarly, following two *events* will be treated the same
```
timbray, line 709: Haha, when I was thinking about this I realized I didn't know what would actually happen.  I'm going to assume you tested this and verified that the first event will *not* match 
```json
{"source": [ "aws.s3"]}
```

Because I'm worried that it might.

src/test/software/amazon/event/ruler/RulerTest.java
```diff
@@ -558,4 +558,88 @@ public void WHEN_WeTryWildcardRules_THEN_TheyWork() throws Exception {
...
+                        "  \"detail\": {\n" +
+                        "      \"eventSource\": [\"s3.amazonaws.com\"]\n" +
+                        "    },\n" +
+                        "  \"detail\": {\n" +
+                        "      \"eventSource\": [\"sns.amazonaws.com\"]\n" +
+                        "    }\n" +
+                        "}"
+        };
+
+        Ruler.matchesRule(eventsWithUniqueKeys, ruleWithUniqueKeys);
```
timbray, line 634: Um, no assertions?

README.md
baldawar: I did test this, just forgot to push the latest commit with the update. Also as the assertions you mentioned below.

README.md
```diff
@@ -679,6 +679,61 @@ It also means that these rules will match against following two events :
...
```
jonessha, line 709: Can be more specific than "treated the same"? How about "will be matched by the same rules"?

README.md
```diff
@@ -679,6 +679,61 @@ It also means that these rules will match against following two events :
...
+{
+  "source": ["aws.sns"],
+  "detail-type": ["AWS API Call via CloudTrail"],
+  "detail": [
+    { "eventSource": ["sns.amazonaws.com"] }
+  ]
+}
+```
+
+This behaviour will change in future version (to avoid any confusions) and should not be relied upon.
```
jonessha, line 735: "Will"? As in we will definitely change the behavior? For backwards compatibility, I'm pretty sure we will keep most of this behavior the same, with the possible exception or rejecting rules (but not events) with duplicate keys. The word "may" is probably more appropriate than "will".

README.md
```diff
@@ -679,6 +679,61 @@ It also means that these rules will match against following two events :
 
 This behaviour may change in future version (to avoid any confusions) and should not be relied upon.
 
+#### Caveat: JSON parsing and duplicate keys
+
+Events or Rules containing duplicate JSON keys are invalid. When duplicate keys are passed Ruler
```
jonessha, line 684: Not really "invalid" because we accept them, right? Maybe just "considered poorly formed"?

src/test/software/amazon/event/ruler/RulerTest.java
```diff
@@ -558,4 +558,103 @@ public void WHEN_WeTryWildcardRules_THEN_TheyWork() throws Exception {
...
+                        "}"
+        };
+
+        assertTrue(Ruler.matchesRule(eventsWithUniqueKeys, ruleWithUniqueKeys));
+        assertFalse(Ruler.matchesRule(eventsWithIgnoredKeys, ruleWithUniqueKeys));
+        assertFalse(Ruler.matchesRule(eventsWithUniqueKeys, ruleWithIgnoredKeys));
+
+        for(String event: eventsWithDuplicateKeys) {
+            assertTrue(Ruler.matchesRule(event, ruleWithUniqueKeys));
+            assertFalse(Ruler.matchesRule(eventsWithIgnoredKeys, ruleWithUniqueKeys));
```
jonessha, line 652: Shouldn't this be outside the loop?

src/test/software/amazon/event/ruler/RulerTest.java
```diff
@@ -558,4 +558,103 @@ public void WHEN_WeTryWildcardRules_THEN_TheyWork() throws Exception {
...
+        }
+
+        for(String rule: rulesWithDuplicateKeys) {
+            assertTrue(Ruler.matchesRule(eventsWithUniqueKeys, rule));
+            assertFalse(Ruler.matchesRule(eventsWithUniqueKeys, ruleWithIgnoredKeys));
```
jonessha, line 657: This too?

PR #39, opened 2022-09-10
some additional performance improvements

rudygt:
### Description of changes:

Following up on the clues from the numeric benchmark slow speed I tested Tim comment about the recursion on ACTask instead of the stepQueue approach (witch is a big issue when we have millions of steps, and the ArrayDeque needs to grow), and indeed we got a nice speed bump by getting rid of the queue and going down using recursion instead. 

there are 3 individual items addresed here 
* remove stepQueue from ACTask and make it recursive instead
* remove Stream.of().collect(Collectors.toSet()) to create a one element sets, that is  very slow compared to Collections.singleton(). ( around 6 times better )
* another small optimization related with getting slices of hex digits used on the Range class ( around 4 times better )

togheter these changes made the numeric benchmark performance go up by around 60%

#### Benchmark / Performance:

```
BEFORE:

Reading citylots2
Read 213068 events
EXACT events/sec: 256708.4
WILDCARD events/sec: 200063.8
PREFIX events/sec: 297581.0
SUFFIX events/sec: 296339.4
EQUALS_IGNORE_CASE events/sec: 256399.5
NUMERIC events/sec: 3476.8
ANYTHING-BUT events/sec: 160806.0
COMBO events/sec: 3466.1
Reading citylots2
Read 213068 events
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Lines: 213068, Msec: 10579
Events/sec: 20140.7
 Rules/sec: 140984.6
Before: 212.3
After: 3029.5
Per rule: -7042
Turning JSON into field-lists...
Finding Rules...
Lines: 213068, Msec: 2055
Events/sec: 103682.7
Before: 216.3
After: 1749.4
Per rule: -3832
Reading lines...
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Lines: 213068, Msec: 1064
Events/sec: 200251.9
 Rules/sec: 729317345.9
Reading citylots2
Read 213068 events
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Matched: 52527
Lines: 213068, Msec: 14695
Events/sec: 14499.4
Reading lines...
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Lines: 213068, Msec: 9711
Events/sec: 21940.9
 Rules/sec: 153586.2


AFTER:

Reading citylots2
Read 213068 events
EXACT events/sec: 273164.1
WILDCARD events/sec: 201007.5
PREFIX events/sec: 307901.7
SUFFIX events/sec: 307901.7
EQUALS_IGNORE_CASE events/sec: 271078.9
NUMERIC events/sec: 5424.3
ANYTHING-BUT events/sec: 172944.8
COMBO events/sec: 5280.0
Reading citylots2
Read 213068 events
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Lines: 213068, Msec: 9846
Events/sec: 21640.1
 Rules/sec: 151480.4
Before: 182.9
After: 3029.5
Per rule: -7116
Turning JSON into field-lists...
Finding Rules...
Lines: 213068, Msec: 2211
Events/sec: 96367.3
Before: 216.3
After: 1749.4
Per rule: -3832
Reading lines...
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Lines: 213068, Msec: 1066
Events/sec: 199876.2
 Rules/sec: 727949020.6
Reading citylots2
Read 213068 events
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Matched: 52527
Lines: 213068, Msec: 10880
Events/sec: 19583.5
Reading lines...
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Lines: 213068, Msec: 9087
Events/sec: 23447.6
 Rules/sec: 164132.9


```

---

### DIFF
--- a/src/main/software/amazon/event/ruler/ByteMachine.java
+++ b/src/main/software/amazon/event/ruler/ByteMachine.java
@@ -12,6 +12,7 @@ import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -812,7 +813,7 @@ class ByteMachine {
         Set<NameState> nextNameStates = new HashSet<>(pattern.getValues().size());
         for (String value : pattern.getValues()) {
             NameState matchPattern = findMatchPattern(getParser().parse(pattern.type(), value), pattern);
-            if (matchPattern != null) {
+            if (Objects.nonNull(matchPattern)) {
                 nextNameStates.add(matchPattern);
             }
         }
--- a/src/main/software/amazon/event/ruler/ByteMachine.java
+++ b/src/main/software/amazon/event/ruler/ByteMachine.java
@@ -12,7 +12,6 @@ import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -810,9 +809,13 @@ class ByteMachine {
 
     private NameState findAnythingButPattern(AnythingBut pattern) {
 
-        Set<NameState> nextNameStates = pattern.getValues().stream().
-                map(value -> findMatchPattern(getParser().parse(pattern.type(), value), pattern)).
-                filter(Objects::nonNull).collect(Collectors.toSet());
+        Set<NameState> nextNameStates = new HashSet<>(pattern.getValues().size());
+        for (String value : pattern.getValues()) {
+            NameState matchPattern = findMatchPattern(getParser().parse(pattern.type(), value), pattern);
+            if (matchPattern != null) {
+                nextNameStates.add(matchPattern);
+            }
+        }
         if (!nextNameStates.isEmpty()) {
             assert nextNameStates.size() == 1 : "nextNameStates.size() == 1";
             return nextNameStates.iterator().next();
--- a/src/main/software/amazon/event/ruler/ByteMachine.java
+++ b/src/main/software/amazon/event/ruler/ByteMachine.java
@@ -12,7 +12,6 @@ import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -813,7 +812,7 @@ class ByteMachine {
         Set<NameState> nextNameStates = new HashSet<>(pattern.getValues().size());
         for (String value : pattern.getValues()) {
             NameState matchPattern = findMatchPattern(getParser().parse(pattern.type(), value), pattern);
-            if (Objects.nonNull(matchPattern)) {
+            if (matchPattern != null) {
                 nextNameStates.add(matchPattern);
             }
         }
--- a/src/main/software/amazon/event/ruler/ByteMap.java
+++ b/src/main/software/amazon/event/ruler/ByteMap.java
@@ -4,7 +4,6 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.NavigableMap;
-import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
 
@@ -194,7 +193,7 @@ class ByteMap {
     Set<ByteTransition> getTransitions() {
         Set<ByteTransition> result = new HashSet<>(map.values().size());
         for (ByteTransition transition : map.values()) {
-            if (Objects.nonNull(transition)) {
+            if (transition != null) {
                 result.add(transition);
             }
         }
--- a/src/main/software/amazon/event/ruler/ByteMap.java
+++ b/src/main/software/amazon/event/ruler/ByteMap.java
@@ -4,10 +4,8 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.NavigableMap;
-import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.stream.Collectors;
 
 import static software.amazon.event.ruler.CompoundByteTransition.coalesce;
 
@@ -193,7 +191,9 @@ class ByteMap {
      * @return All transitions contained in this map.
      */
     Set<ByteTransition> getTransitions() {
-        return map.values().stream().filter(Objects::nonNull).collect(Collectors.toSet());
+        Set<ByteTransition> result = new HashSet<>(map.values());
+        result.remove(null);
+        return result;
     }
 
     /**
--- a/src/main/software/amazon/event/ruler/ByteMap.java
+++ b/src/main/software/amazon/event/ruler/ByteMap.java
@@ -4,6 +4,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.NavigableMap;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
 
@@ -191,8 +192,12 @@ class ByteMap {
      * @return All transitions contained in this map.
      */
     Set<ByteTransition> getTransitions() {
-        Set<ByteTransition> result = new HashSet<>(map.values());
-        result.remove(null);
+        Set<ByteTransition> result = new HashSet<>(map.values().size());
+        for (ByteTransition transition : map.values()) {
+            if (Objects.nonNull(transition)) {
+                result.add(transition);
+            }
+        }
         return result;
     }
 
--- a/src/main/software/amazon/event/ruler/Range.java
+++ b/src/main/software/amazon/event/ruler/Range.java
@@ -1,9 +1,7 @@
 package software.amazon.event.ruler;
 
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
 
 /**
  * Represents a range of numeric values to match against.
@@ -56,20 +54,26 @@ public class Range extends Patterns {
     public static Range lessThan(final double val) {
         return new Range(-Constants.FIVE_BILLION, false, val, true);
     }
+
     public static Range lessThanOrEqualTo(final double val) {
         return new Range(-Constants.FIVE_BILLION, false, val, false);
     }
+
     public static Range greaterThan(final double val) {
         return new Range(val, true, Constants.FIVE_BILLION, false);
     }
+
     public static Range greaterThanOrEqualTo(final double val) {
         return new Range(val, false, Constants.FIVE_BILLION, false);
     }
+
     public static Range between(final double bottom, final boolean openBottom, final double top, final boolean openTop) {
         return new Range(bottom, openBottom, top, openTop);
     }
 
-    private static Range deepCopy(final Range range) { return new Range(range); }
+    private static Range deepCopy(final Range range) {
+        return new Range(range);
+    }
 
     /**
      * This is necessitated by the fact that we do range comparisons of numbers, fixed-length strings of digits, and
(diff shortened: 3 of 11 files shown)


baldawar:
Will look at this on Monday Pacific time. Just dropped in to make some minor callouts. 
1. You don't need to use git force push unless you like using it. When merging a pull-request, we squash everything into a single commit. 
2. We should start updating the README for performance gains and do a minor version bump in pom.xml 
3. This made my weekend even better. Thank you. ❤️

timbray:
Before looking at the code:

1. When I originally built the StepQueue implementation, my idea was that this was leaving the door open so that the step/field matching tasks could be dealt out to multiple threads. Which isn't going to be possible with the recursive approach. But then I decided that going concurrent probably wouldn't be a win because each field/state match is a tiny amount of work, the inevitable cost of synchronization would probably kill the benefit. But, I never tried it.
2. Could you do a quickie measure of the recursion depth?  The JVM stack size is typically about 1M/thread. There's a reason that recursive code is strongly disapproved of in kernel code; I'm wondering if something like CityLots, only more so, could blow up the recursive approach.

src/main/software/amazon/event/ruler/ByteMap.java
```diff
@@ -193,7 +191,9 @@ ByteTransition getTransitionForAllBytes() {
      * @return All transitions contained in this map.
      */
     Set<ByteTransition> getTransitions() {
-        return map.values().stream().filter(Objects::nonNull).collect(Collectors.toSet());
+        Set<ByteTransition> result = new HashSet<>(map.values());
+        result.remove(null);
```
timbray, line 195: Huh? Don't know this idiom, why `remove(null)`?

src/test/software/amazon/event/ruler/CIDRTest.java
```diff
@@ -61,39 +61,40 @@ public void testToStringFailures() {
     @Test
     public void TestDigitSequence() {
 
-        List<Byte> l = Range.digitSequence((byte) '4', (byte) 'C', false, false);
-        byte[] wanted = { '5', '6', '7', '8', '9', 'A', 'B' };
+        byte[] l = Range.digitSequence((byte) '4', (byte) 'C', false, false);
+        byte[] wanted = {'5', '6', '7', '8', '9', 'A', 'B'};
         for (int i = 0; i < wanted.length; i++) {
-            assertEquals(wanted[i], (byte) l.get(i));
```
timbray, line 67: Since I work mostly in Go these days, I really like absolutely unified formatting.  What tool are you using, and maybe we should put in a GitHub action or something to iron out the formatting.  In the IntelliJ Go IDE it now runs gofmt on every save :)

timbray (commented):
will need to check out your branch to really understand the new ACFinder implementation. In the back of my mind I'm wondering if, after removing the stepQueue, we still need the ACTask and ACStep classes.  Will pull branch and have a look.

src/main/software/amazon/event/ruler/ByteMap.java
rudygt: basically stream().filter().collect() is creating a set from the values removing all nulls in the process. but there is some overhead going that route (stream is slow in hot paths), so I basically tried to reproduce the same that why I remove null after creating the set, as far as i remember the set can hold one copy of null, I will double check to make sure this is needed.

src/test/software/amazon/event/ruler/CIDRTest.java
rudygt: intellij idea with default format, but yep I have notice very small differences vs the format, some unintentional reformats still slipping tho jaja, agree that would be nice to have a format enforced by the project so we dont have to care about them

rudygt:
> Before looking at the code:
> 
> 1. When I originally built the StepQueue implementation, my idea was that this was leaving the door open so that the step/field matching tasks could be dealt out to multiple threads. Which isn't going to be possible with the recursive approach. But then I decided that going concurrent probably wouldn't be a win because each field/state match is a tiny amount of work, the inevitable cost of synchronization would probably kill the benefit. But, I never tried it.
> 2. Could you do a quickie measure of the recursion depth?  The JVM stack size is typically about 1M/thread. There's a reason that recursive code is strongly disapproved of in kernel code; I'm wondering if something like CityLots, only more so, could blow up the recursive approach.

I did check on the deep level and it stays at 4 on the benchmarks, 8 in ACMachineTest, and 9 across the whole test suite, We should probably add one specific test to stress deeper that that.

baldawar:
Here's a dummy JSON I've used in the past [deep_json.txt](https://github.com/aws/event-ruler/files/9550283/deep_json.txt). Probably can building some programmatically and add a benchmark test around it.

src/main/software/amazon/event/ruler/ByteMachine.java
```diff
@@ -810,9 +809,13 @@ NameState findPattern(final Patterns pattern) {
...
-        Set<NameState> nextNameStates = pattern.getValues().stream().
-                map(value -> findMatchPattern(getParser().parse(pattern.type(), value), pattern)).
-                filter(Objects::nonNull).collect(Collectors.toSet());
+        Set<NameState> nextNameStates = new HashSet<>(pattern.getValues().size());
+        for (String value : pattern.getValues()) {
+            NameState matchPattern = findMatchPattern(getParser().parse(pattern.type(), value), pattern);
+            if (matchPattern != null) {
+                nextNameStates.add(matchPattern);
+            }
+        }
```
baldawar, line 818: super-duper nitpicking, but I think you can have a consistent style here and in `ByteMap.getTransitions()` if you for `x.forEach( elem -> addWhenNotNull (...) )`. I'm not sure on the performance implications yet but maybe slightly readable. Maybe...

src/main/software/amazon/event/ruler/ByteMap.java
baldawar: If you end up keeping this, would recommend a short comment or test on why. I can see someone missing this in a refactor fairly easily.

src/test/software/amazon/event/ruler/CIDRTest.java
baldawar: Added an issue for this https://github.com/aws/event-ruler/issues/43

baldawar (changes requested):
Gentle reminder to update the readme and pom.xml in the next iteration.

src/main/software/amazon/event/ruler/ByteMap.java
rudygt: I went for having a consistent style on those two, I believe it is easy to understand now.

src/main/software/amazon/event/ruler/ByteMachine.java
rudygt: I did update this one to the new style that I believe is easier to understand.

rudygt:
> Gentle reminder to update the readme and pom.xml in the next iteration.

I have increased the minor version on the pom.

not sure what to do about readme performance section, my machine is a desktop Ryzen9 5900x so my numbers look very different than the reference macbook, maybe you Rishi can take care of updating the numbers after your benchmark split lands on master ?

I also added a new test to make sure deep nested events and rules dont break anything

Thanks!.

src/main/software/amazon/event/ruler/ByteMachine.java
```diff
@@ -810,9 +810,13 @@ NameState findPattern(final Patterns pattern) {
 
     private NameState findAnythingButPattern(AnythingBut pattern) {
 
-        Set<NameState> nextNameStates = pattern.getValues().stream().
-                map(value -> findMatchPattern(getParser().parse(pattern.type(), value), pattern)).
-                filter(Objects::nonNull).collect(Collectors.toSet());
+        Set<NameState> nextNameStates = new HashSet<>(pattern.getValues().size());
+        for (String value : pattern.getValues()) {
+            NameState matchPattern = findMatchPattern(getParser().parse(pattern.type(), value), pattern);
+            if (Objects.nonNull(matchPattern)) {
```
timbray, line 816: Um, the javadocs say that `Objects.nonNull()` is designed to be used while filtering.  Obviously it does no harm, but when I saw this I had to go read a bunch of stuff to figure out whether this was better for some cosmic reason than just `matchPattern != null` and it's not.  So maybe let's not make other people get the same puzzled look on their face and and up at StackOverflow…

timbray (commented):
Aside from one fairly trivial complaint, I'm good on this PR.

src/main/software/amazon/event/ruler/ByteMachine.java
rudygt: jajaja, I got biased by the streams (very common in that context) , I did try to make it closer to the other side, but I think you are right,  != null is crystal clear

timbray (approved):

src/main/software/amazon/event/ruler/ByteMap.java
```diff
@@ -193,7 +191,13 @@ ByteTransition getTransitionForAllBytes() {
      * @return All transitions contained in this map.
      */
     Set<ByteTransition> getTransitions() {
-        return map.values().stream().filter(Objects::nonNull).collect(Collectors.toSet());
+        Set<ByteTransition> result = new HashSet<>(map.values().size());
```
jonessha, line 194: Just a note that the Set may end up containing less elements than map.values().size(), because 1) nulls are removed, and 2) the values in the map are not unique. I suppose it's probably more efficient to oversize the Set initially though than to undersize it by going with the default constructor.

src/main/software/amazon/event/ruler/Range.java
```diff
@@ -56,53 +54,67 @@ private Range(Range range) {
...
         return bytes;
     }
 
+    private static int getHexByteIndex(byte value) {
+        // ['0'-'9'] maps to [0-9] indexes
+        if (value >= 48 && value <= 57) {
+            return value - 48;
+        }
+        // ['A'-'F'] maps to [10-15] indexes
+        return (value - 65) + 10;
```
jonessha, line 115: Any reason not to just go with -55 here? If you're trying to lay out both steps here for understandability/readability, perhaps it would be better to go further and turn these numbers into private static final class variables?

src/main/software/amazon/event/ruler/ByteMap.java
rudygt: yes good observation, I was aware of this, I just took the size of the values as the upper bound of the set size, but perf wise its cheaper to allocate above, than allowing the set to grow

jonessha (approved):
LGTM, thanks for the changes!

src/main/software/amazon/event/ruler/Range.java
rudygt: mostly to try to keep it easy to understand, if we trust the compiler this is being optimized away. 

I think the variable may be the better option then

src/main/software/amazon/event/ruler/Range.java
```diff
@@ -108,11 +109,11 @@ static byte[] digitSequence(byte first, byte last, boolean includeFirst, boolean
 
     private static int getHexByteIndex(byte value) {
```
rudygt, line 110: @jonessha what do you think? I feel like using the chars helps with the clarity , and then only one obscure constant is needed

baldawar (approved):

baldawar:
Thanks again @rudygt

src/main/software/amazon/event/ruler/Range.java
jonessha: This looks good! Thank you!

PR #58, opened 2022-11-14
Adding back try/catch inadvertently removed from CIDR code

jonessha:
This try/catch was inadvertently removed as part of the wildcard matching PR (#18). The effect of this is certain strings that fail to be parsed as CIDR will bubble-up an exception instead of being treated as an exact match. Reverting to be like before.


baldawar:
Is it worth adding a unit test that confirms the behaviour?

baldawar (approved):

Issue #43, opened 2022-09-12
Better / Consistent Formatting


baldawar:
## What is your idea?

Add some means to have consistent formatting across the codebase. See comment https://github.com/aws/event-ruler/pull/39/files#r967886681 for some early thoughts. 

## Would you be willing to make the change? 

Maybe

baldawar:
I looked into this a bit more while reviewing https://github.com/aws/event-ruler/pull/60. Seems there's a bunch of options listed within https://github.com/google/google-java-format. Love the idea of a github actions so that we don't have to talk about formatting during reviews. 

There's also https://github.com/revelc/formatter-maven-plugin. 

All of these are tools I'm not too familiar with, so more deep dive is required on if there's any gotchas or regressions that such tools can introduce.

schenksj:
Might not be a terrible idea to add `husky` (https://github.com/typicode/husky) hooks into the repo to assure that code has been formatted, along with a 1-time reformat PR for the selected formatter.  Only real downside is that husky will required you to install node/npm on any machine used for development.

My teams at my day job have had great success with enforcing this in python with `black`, `pylint`, and `pycov` (and for golang `go test -cover`, `go fmt`, and `golangci-lint`).  I believe similar open source tools exist for java as well.  Personally, I've had great luck with google-java-format just using the default configuration.  It is wonderful how (whether you like how the resulting code looks or not), there is a single correct answer to things that matter, but only sort of.

PR #60, opened 2022-11-18
enhancement refactor Utility class Constants, Task;

h1alexbel:
Small Style Refactor for Constants Utility class, and Task class.

### DIFF
--- a/src/main/software/amazon/event/ruler/Constants.java
+++ b/src/main/software/amazon/event/ruler/Constants.java
@@ -4,60 +4,62 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.regex.Pattern;
 
-class Constants {
+final class Constants {
 
-    private Constants() { }
+  private Constants() {
+    throw new UnsupportedOperationException();
+  }
 
-    final static String EXACT_MATCH = "exactly";
-    final static String EQUALS_IGNORE_CASE = "equals-ignore-case";
-    final static String PREFIX_MATCH = "prefix";
-    final static String SUFFIX_MATCH = "suffix";
-    final static String ANYTHING_BUT_MATCH = "anything-but";
-    final static String EXISTS_MATCH = "exists";
-    final static String WILDCARD = "wildcard";
-    final static String NUMERIC = "numeric";
-    final static String CIDR = "cidr";
+  static final String EXACT_MATCH = "exactly";
+  static final String EQUALS_IGNORE_CASE = "equals-ignore-case";
+  static final String PREFIX_MATCH = "prefix";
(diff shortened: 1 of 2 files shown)


src/main/software/amazon/event/ruler/Task.java
```diff
@@ -46,39 +47,39 @@ class Task {
     }
 
     NameState startState() {
-        return machine.getStartState();
+        return this.machine.getStartState();
```
timbray, line 50: Why? (recurs)

src/main/software/amazon/event/ruler/Constants.java
```diff
@@ -4,60 +4,62 @@
...
+final class Constants {
+
+  private Constants() {
+    throw new UnsupportedOperationException();
+  }
+
+  static final String EXACT_MATCH = "exactly";
+  static final String EQUALS_IGNORE_CASE = "equals-ignore-case";
+  static final String PREFIX_MATCH = "prefix";
+  static final String SUFFIX_MATCH = "suffix";
```
baldawar, line 16: Looks like standard across this pkg is "final static" https://github.com/aws/event-ruler/search?q=static+final, so changing this in just one file would make it inconsistent. In general, I'd favour consistency over ideal practices in just one part of the code.

baldawar:
Appreciate the effort @h1alexbel  but I'm hesitant to take this one in because its not consistent across the pkg. Personally, I'd love for someone to take on https://github.com/aws/event-ruler/issues/43.

src/main/software/amazon/event/ruler/Constants.java
timbray: > Looks like standard across this pkg is "final static

That would be my fault. Oops.

h1alexbel: Hm, if its not expected formatting, can I change all ``` final static ``` to ``` static final```?
@timbray @baldawar

h1alexbel:
@baldawar updated PR with the same pkg ```final + static``` policy as you have. Please take a look.

src/main/software/amazon/event/ruler/Task.java
h1alexbel: using ```this.field``` gives more understanding that this field is related to the class. The same with methods.

baldawar: I'd recommend to not change this here as nearly all other classes aren't using `this.`

I want the codebase to stay consistent first, and not be deviate unless there's a code reason for one class to be unique on its own.

h1alexbel:
this is the duplication of #61

PR #61, opened 2022-11-21
Input refactor

h1alexbel:
Refactored ```software.amazon.event.ruler.input``` package. Introducing new interfaces for existing classes to decouple them.

### DIFF
--- a/src/main/software/amazon/event/ruler/Constants.java
+++ b/src/main/software/amazon/event/ruler/Constants.java
@@ -4,60 +4,62 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.regex.Pattern;
 
-class Constants {
+final class Constants {
 
-    private Constants() { }
+  private Constants() {
+    throw new UnsupportedOperationException();
+  }
 
-    final static String EXACT_MATCH = "exactly";
-    final static String EQUALS_IGNORE_CASE = "equals-ignore-case";
-    final static String PREFIX_MATCH = "prefix";
-    final static String SUFFIX_MATCH = "suffix";
-    final static String ANYTHING_BUT_MATCH = "anything-but";
-    final static String EXISTS_MATCH = "exists";
-    final static String WILDCARD = "wildcard";
-    final static String NUMERIC = "numeric";
-    final static String CIDR = "cidr";
+  static final String EXACT_MATCH = "exactly";
+  static final String EQUALS_IGNORE_CASE = "equals-ignore-case";
+  static final String PREFIX_MATCH = "prefix";
+  static final String SUFFIX_MATCH = "suffix";
+  static final String ANYTHING_BUT_MATCH = "anything-but";
+  static final String EXISTS_MATCH = "exists";
+  static final String WILDCARD = "wildcard";
+  static final String NUMERIC = "numeric";
+  static final String CIDR = "cidr";
 
-    // This is Ruler reserved words to represent the $or relationship among the fields.
-    final static String OR_RELATIONSHIP_KEYWORD = "$or";
+  // This is Ruler reserved words to represent the $or relationship among the fields.
+  static final String OR_RELATIONSHIP_KEYWORD = "$or";
 
-    final static String EQ = "=";
-    final static String LT = "<";
-    final static String LE = "<=";
-    final static String GT = ">";
-    final static String GE = ">=";
+  static final String EQ = "=";
+  static final String LT = "<";
+  static final String LE = "<=";
+  static final String GT = ">";
+  static final String GE = ">=";
 
-    // Use scientific notation to define the double number directly to avoid losing Precision by calculation
-    // for example 5000 * 1000 *1000 will be wrongly parsed as 7.05032704E8 by computer.
-    final static double FIVE_BILLION = 5E9;
+  // Use scientific notation to define the double number directly to avoid losing Precision by calculation
+  // for example 5000 * 1000 *1000 will be wrongly parsed as 7.05032704E8 by computer.
+  static final double FIVE_BILLION = 5E9;
 
-    final static Pattern IPv4_REGEX = Pattern.compile("[0-9]+\\.[0-9]+\\.[0-9]+\\.[0-9]+");
-    final static Pattern IPv6_REGEX = Pattern.compile("[0-9a-fA-F:]*:[0-9a-fA-F:]*");
-    final static byte[] HEX_DIGITS = {
-            '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
-            'A', 'B', 'C', 'D', 'E', 'F'
-    };
-    final static byte MAX_DIGIT = 'F';
+  static final Pattern IPv4_REGEX = Pattern.compile("[0-9]+\\.[0-9]+\\.[0-9]+\\.[0-9]+");
+  static final Pattern IPv6_REGEX = Pattern.compile("[0-9a-fA-F:]*:[0-9a-fA-F:]*");
+  static final byte[] HEX_DIGITS = {
+      '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
+      'A', 'B', 'C', 'D', 'E', 'F'
+  };
+  static final byte MAX_DIGIT = 'F';
 
-    final static List<String> RESERVED_FIELD_NAMES_IN_OR_RELATIONSHIP = Arrays.asList(
-            EXACT_MATCH,
-            EQUALS_IGNORE_CASE,
-            PREFIX_MATCH,
-            SUFFIX_MATCH,
-            ANYTHING_BUT_MATCH,
-            EXISTS_MATCH,
-            WILDCARD,
-            NUMERIC,
-            CIDR,
-            // Numeric comparisons
-            EQ, LT, LE, GT, GE,
-            // reserve below keywords for future extension
-            "regex",
-            // String Comparisons
-            "not-wildcard", "not-equals-ignore-case",
-            // Date/Time comparisons
-            "date-after", "date-on-or-after", "date-before", "date-on-or-before", "in-date-range",
-            // IP Address Comparison
-            "ip-address-in-range", "ip-address-not-in-range"
-    );
+  static final List<String> RESERVED_FIELD_NAMES_IN_OR_RELATIONSHIP = Arrays.asList(
+      EXACT_MATCH,
+      EQUALS_IGNORE_CASE,
+      PREFIX_MATCH,
+      SUFFIX_MATCH,
+      ANYTHING_BUT_MATCH,
+      EXISTS_MATCH,
+      WILDCARD,
+      NUMERIC,
+      CIDR,
+      // Numeric comparisons
+      EQ, LT, LE, GT, GE,
+      // reserve below keywords for future extension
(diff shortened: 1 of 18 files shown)


src/main/software/amazon/event/ruler/input/MatchTypeParser.java
```diff
@@ -0,0 +1,17 @@
+package software.amazon.event.ruler.input;
+
+import software.amazon.event.ruler.MatchType;
+
+/**
+ * @author Aliaksei Bialiauski (abialiauski@solvd.com)
```
baldawar, line 6: please remove

src/main/software/amazon/event/ruler/input/DefaultParser.java
```diff
@@ -12,7 +12,8 @@
  * rule's value will be treated by their byte representation, but certain characters, such as for wildcards or regexes,
  * need to be represented differently so the Machine understands their special meaning.
  */
-public class Parser {
+@SuppressWarnings("UseOfConcreteClass")
+public class DefaultParser implements MatchTypeParser, ByteParser {
```
baldawar, line 16: I don't think the renaming is necessary but if you are doing so, please check if there's any documentation (in code or Readme) that need to be updated as well. I've found IDEs to not be able to pick them all.

src/main/software/amazon/event/ruler/input/WildcardParser.java
```diff
@@ -3,30 +3,33 @@
...
     WildcardParser() { }
 
-    public InputCharacter[] parse(String value) {
-        byte[] utf8bytes = value.getBytes(StandardCharsets.UTF_8);
-        List<InputCharacter> result = new ArrayList<>(utf8bytes.length);
-        for (int i = 0; i < utf8bytes.length; i++) {
-            byte utf8byte = utf8bytes[i];
+    @Override
+    public InputCharacter[] parse(final MatchType type, final String value) {
+        final byte[] utf8 = value.getBytes(StandardCharsets.UTF_8);
```
baldawar, line 21: utf8Bytes is more accurate here.

src/main/software/amazon/event/ruler/input/WildcardParser.java
```diff
@@ -3,30 +3,33 @@
...
-import static software.amazon.event.ruler.input.Parser.ASTERISK_BYTE;
-import static software.amazon.event.ruler.input.Parser.BACKSLASH_BYTE;
+import static software.amazon.event.ruler.input.DefaultParser.ASTERISK_BYTE;
+import static software.amazon.event.ruler.input.DefaultParser.BACKSLASH_BYTE;
 
 /**
  * A parser to be used specifically for wildcard rules.
  */
-public class WildcardParser {
+@SuppressWarnings({"UseOfConcreteClass", "ObjectAllocationInLoop"})
```
baldawar, line 14: curious if you've considered suppressing individual methods first? Ideally you want to suppress the smallest part of the code (and add a comment on why suppression is necessary).

src/main/software/amazon/event/ruler/input/EqualsIgnoreCaseParser.java
```diff
@@ -13,20 +14,22 @@
...
  */
-public class EqualsIgnoreCaseParser {
+@SuppressWarnings({"UseOfConcreteClass", "ObjectAllocationInLoop",
+    "CollectionWithoutInitialCapacity"})
+public class EqualsIgnoreCaseParser implements StringValueParser {
 
     EqualsIgnoreCaseParser() { }
 
-    public InputCharacter[] parse(String value) {
+    public InputCharacter[] parse(final MatchType type, final String value) {
```
baldawar, line 23: is `type` redundant here?

src/main/software/amazon/event/ruler/MatchType.java
```diff
@@ -14,5 +14,6 @@ public enum MatchType {
     ANYTHING_BUT,        // black list effect
     ANYTHING_BUT_PREFIX, // anything that doesn't start with this
     EQUALS_IGNORE_CASE,  // case-insensitive string match
-    WILDCARD             // string match using one or more non-consecutive '*' wildcard characters
-}
+    WILDCARD,             // string match using one or more non-consecutive '*' wildcard characters
+    NO
```
baldawar, line 18: Can you elaborate why we have `NO` here?

src/main/software/amazon/event/ruler/Constants.java
```diff
@@ -4,60 +4,62 @@
...
-            // Date/Time comparisons
-            "date-after", "date-on-or-after", "date-before", "date-on-or-before", "in-date-range",
-            // IP Address Comparison
-            "ip-address-in-range", "ip-address-not-in-range"
-    );
-}
+final class Constants {
+
+  private Constants() {
+    throw new UnsupportedOperationException();
```
baldawar, line 10: I recommend whenever you throw an exception to also include a contextual message.

src/main/software/amazon/event/ruler/Task.java
```diff
@@ -46,39 +47,39 @@ class Task {
...
     // The field used means all steps in the field must be all used individually.
     boolean isFieldUsed(final String field) {
         if (field.contains(".")) {
-            String[] steps = field.split("\\.");
-            return Arrays.stream(steps).allMatch(machine::isFieldStepUsed);
+            final String[] steps = field.split("\\.");
+            return Arrays.stream(steps).allMatch(this.machine::isFieldStepUsed);
         }
-        return machine.isFieldStepUsed(field);
+        return this.machine.isFieldStepUsed(field);
```
baldawar, line 59: is this duplicate of https://github.com/aws/event-ruler/pull/60/files

src/main/software/amazon/event/ruler/input/StringValueParser.java
```diff
@@ -0,0 +1,14 @@
...
+import software.amazon.event.ruler.MatchType;
+
+/**
+ * @author Aliaksei Bialiauski (abialiauski@solvd.com)
+ */
+@SuppressWarnings("UseOfConcreteClass")
+public interface StringValueParser extends MatchTypeParser {
+
+  default InputCharacter[] parse(final String value) {
+    return this.parse(MatchType.NO, value);
```
baldawar, line 12: Needs documentation (and supporting unit tests) to understand the context and expected behavior.

src/main/software/amazon/event/ruler/MatchType.java
h1alexbel: Yeah, because in Parser's method #parse(MatchType type, String value) we need to pass MatchType, to prevent using [```null```](https://www.yegor256.com/2014/05/13/why-null-is-bad.html), I suggest to use [Null Object Pattern](https://www.baeldung.com/java-null-object-pattern) instead.

h1alexbel: > Can you elaborate why we have `NO` here?

Yeah, because in Parser's method #parse(MatchType type, String value) we need to pass MatchType, to prevent using [```null```](https://www.yegor256.com/2014/05/13/why-null-is-bad.html), I suggest to use [Null Object Pattern](https://www.baeldung.com/java-null-object-pattern) instead.

src/main/software/amazon/event/ruler/input/WildcardParser.java
h1alexbel: My bad, I'll remove that.

h1alexbel:
Hey, @baldawar I've updated PR with the consistent pkg code style as you have. Please take a look.

src/main/software/amazon/event/ruler/MatchType.java
```diff
@@ -14,5 +14,6 @@ public enum MatchType {
     ANYTHING_BUT,        // black list effect
     ANYTHING_BUT_PREFIX, // anything that doesn't start with this
     EQUALS_IGNORE_CASE,  // case-insensitive string match
-    WILDCARD             // string match using one or more non-consecutive '*' wildcard characters
-}
+    WILDCARD,             // string match using one or more non-consecutive '*' wildcard characters
+    NO // needs for null object pattern
```
baldawar, line 18: nit, formatting of comment is off.

src/main/software/amazon/event/ruler/RuleCompiler.java
```diff
@@ -28,11 +28,13 @@
...
  * Is public so clients can call the check() method to syntax-check filters
  */
-public class RuleCompiler {
+public final class RuleCompiler {
 
     private static final JsonFactory JSON_FACTORY = new JsonFactory();
 
-    private RuleCompiler() { }
+    private RuleCompiler() {
+      throw new UnsupportedOperationException();
```
baldawar, line 36: missed a message here.

src/main/software/amazon/event/ruler/input/ByteParser.java
```diff
@@ -0,0 +1,10 @@
+package software.amazon.event.ruler.input;
+
+public interface ByteParser {
```
baldawar, line 3: I'd recommend you also include a minor doc on what we expect to use this interface. nothing too fancy, but enough to offer direction for anyone that comes across this class.

src/main/software/amazon/event/ruler/input/StringValueParser.java
```diff
@@ -0,0 +1,10 @@
+package software.amazon.event.ruler.input;
+
+public interface StringValueParser {
```
baldawar, line 3: I think I need one more input on the inclusion of these parsers. @jonessha  do you have an opinions on either keeping / removing / changing interfaces?

baldawar:
Thanks @h1alexbel . Mostly LGTM. I just want a second opinion on the interfaces from Shawn as it might affect something he's potentially working in recent future. If he has no objections by tomorrow, probably good to merge if you can address the minor call-outs I made.

src/main/software/amazon/event/ruler/input/StringValueParser.java
jonessha: I'm not strongly opposed to this change, but at the same time, I don't really get the purpose of including all these interfaces. In my mind for how Ruler will evolve, these interfaces will serve no purpose. @h1alexbel - curious what benefit you see these interfaces providing as the code base evolves?

h1alexbel: > I'm not strongly opposed to this change, but at the same time, I don't really get the purpose of including all these interfaces. In my mind for how Ruler will evolve, these interfaces will serve no purpose. @h1alexbel - curious what benefit you see these interfaces providing as the code base evolves?

Benefit is maintainability in the future and possible API extension via interfaces, and that all classes will be used by some special contract.

baldawar: looks like this PR has stalled. since there's no blockers with any changes Shawn has staged in a different branch / repo; I don't mind merging this for now. We can always undo if we think this creates too much friction in future.

baldawar (approved):

baldawar:
@h1alexbel  thanks for the PR. 

If you end up wanting to make more changes in future; if I can get an heads up, I'll make sure to keep some capacity on the side for your changes.

PR #75, opened 2023-02-17
Exponential memory improvement by re-using NameState across multiple patterns

jonessha:
### Description of changes:
Seeking feedback from Tim and Long. This seems like a huge improvement to the fundamental Ruler code for minimal effort, so it has occurred to me that I might be overlooking something. I'd love some feedback about the validity of the approach. (It seems valid in my mind.)

Previously, Ruler would create a new NameState for every pattern. Now, NameStates are re-used across patterns for a given field.

The NameState is accessed via the pattern's ByteMatch at the end of the ByteMachine and provides access to the ByteMachine for the next field. So picture a rule with many fields, and for each of these fields, an array of many elements. Each array element is a pattern for the field. So this turns into a rapidly/exponentially expanding tree of NameStates and ByteMachines. But it's all unnecessary. Since the next field and its patterns will be the same no matter what value we use to match the current field, there is no reason not to re-use NameStates (and thus ByteMachines). This collapses an exponentially-expanding tree into a line. Or to word it another way, this changes Ruler's memory complexity from O(product of all array sizes) to O(sum of all array sizes).

Testing against one event with several large arrays, there was a 99.5% reduction in the number of ByteMatches created.

---

### DIFF
--- a/src/main/software/amazon/event/ruler/ByteMachine.java
+++ b/src/main/software/amazon/event/ruler/ByteMachine.java
@@ -624,18 +624,28 @@ class ByteMachine {
         return false;
     }
 
-    // Adds one pattern to a byte machine.
     NameState addPattern(final Patterns pattern) {
+        return addPattern(pattern, null);
+    }
+
+    /**
+     * Adds one pattern to a byte machine.
+     *
+     * @param pattern The pattern to add.
+     * @param nameState If non-null, transition to this NameState from the ByteMatch.
+     * @return NameState transitioned to from ByteMatch. Will be equal to provided NameState if it wasn't null.
+     */
+    NameState addPattern(final Patterns pattern, final NameState nameState) {
         switch (pattern.type()) {
             case NUMERIC_RANGE:
                 assert pattern instanceof Range;
-                return addRangePattern((Range) pattern);
+                return addRangePattern((Range) pattern, nameState);
             case ANYTHING_BUT:
                 assert pattern instanceof AnythingBut;
-                return addAnythingButPattern((AnythingBut) pattern);
+                return addAnythingButPattern((AnythingBut) pattern, nameState);
             case ANYTHING_BUT_IGNORE_CASE:
                 assert pattern instanceof AnythingButEqualsIgnoreCase;
-                return addAnythingButEqualsIgnoreCasePattern((AnythingButEqualsIgnoreCase) pattern);
+                return addAnythingButEqualsIgnoreCasePattern((AnythingButEqualsIgnoreCase) pattern, nameState);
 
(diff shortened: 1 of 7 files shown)


timbray:
I will review this, but first: I'm inclined to believe you're right, that Ruler creates a lot more states than it needs to. The fact that tests pass is evidence in your favor.  Every time I think about this I get a headache, but my thoughts have been something along the line of "there should be a pool of name states with one entry for every field-name/matched-value-pattern." Or maybe "…for every unique path through the machine".

I wrote the following note to myself in the Quamina code, which may be relevant. In Quamina speak, a "fieldMatcher" is a Ruler NameState, a "pattern" is a Rule.
```go
	// suppose I'm adding the first pattern to a matcher and it has "x": [1, 2]. In principle the branches on
	//  "x": 1 and "x": 2 could go to tne same next state. But we have to make a unique next state for each of them
	//  because some future other pattern might have "x": [2, 3] and thus we need a separate branch to potentially
	//  match two patterns on "x": 2 but not "x": 1. If you were optimizing the automaton for size you might detect
	//  cases where this doesn't happen and reduce the number of fieldMatchers
```

jonessha:
Thanks for the fast response, @timbray. You're giving me some reassurance this should work as I expect. Your Quamina note was useful in checking my thinking. I addressed this by storing both the rule name and the pattern inside NameState. This allows us to delete one rule & pattern, but leave the branch in place if another rule & pattern is still using it.

timbray:
A quick glance at the code didn't turn up any measurements of the savings? Maybe I just missed it?  I saw the NameState.toString(), does this turn up any useful data on the savings?

jonessha:
@timbray Rishi has a separate change in-flight that provides a means to measure a Machine's size. Once that is merged as well, we will see the size of some Machines drop massively in the tests he added. So I didn't add any further tests. But I did some additional manual testing, which showed that for one event with several large arrays, there was a 99.5% reduction in the number of ByteMatches created.

timbray:
Heh, we did Ruler in a big hurry-up rush to ship “CloudWatch Events” and then it worked well enough, so we just kept piling on features. Your typical CW Events customer had like 3 rules each mentioning 1 or 2 fields, outliers had maybe 15… I remember when *redacted* (huge non-AWS service, Long's former home) tried building machines with a million rules and they worked but well yeah, memory pain, then we did all that polymorphic stuff with ByteTransition to save memory… would like to track down that test and try it again.  

Globally across the universe of deployed software, there are low-hanging fruit all over the place that nobody ever thinks to look for.

jonessha:
Closing out this pull request. Opened a new version here: https://github.com/aws/event-ruler/pull/88

Issue #64, opened 2022-11-29
why not make deletePatternRule public ? 


happyomg:
## What is your idea?
i use `addPatternRule` to add `Patterns` in machine, but when i want to delete rule by ruleName,  i found that `deletePatternRule` method is not `public`. 

## Would you be willing to make the change? 

Yes, Just make `deletePatternRule` method `public` is enough.  
If the user can reconstruction the patterns, the `deletePatternRule` method actually don't care about whether the input `Patterns` objects is the same object that added before or not.

i try test in software.amazon.event.ruler.ACMachineTest#dynamicDeleteRules , and the test passed.
```
 @Test
    public void dynamicDeleteRules() throws Exception {
        Machine machine = new Machine();

        TestEvent e = new TestEvent("0", "\"\"", "a", "11", "b", "21", "c", "31", "gamma", "41", "zoo", "\"keeper\"");

        Rule rule;
        Rule rule1;
        rule = new Rule("R1");
        rule.setKeys("a", "b", "c");
        rule.setExactMatchValues("11", "21", "31");
        machine.addPatternRule(rule.name, rule.fields);
        List<String> actual = machine.rulesForJSONEvent(e.toString());
        assertEquals(1, actual.size());
        //System.out.println("matched rule:" + actual);

        rule1 = new Rule("R1");
        rule1.setKeys("a", "b","gamma");
        rule1.setExactMatchValues("11", "21","41");
        machine.addPatternRule(rule1.name, rule1.fields);
        List<String> actual1 = machine.rulesForJSONEvent(e.toString());
        assertEquals(1, actual1.size());
        //System.out.println("matched rule:" + actual1);



        // delete R1 subset with rule.fields
        Rule rule2 = new Rule("R1");
        rule2.setKeys("a", "b", "c");
        rule2.setExactMatchValues("11", "21", "31");
        machine.deletePatternRule(rule2.name, rule2.fields);
        List<String> actual2 = machine.rulesForJSONEvent(e.toString());
        assertEquals(1, actual2.size());
        //System.out.println("matched rule:" + actual2);

        // delete R1 subset with rule1 fields, after this step,
        // the machine will become empty as if no rule was added before.
        Rule rule3 = new Rule("R1");
        rule3.setKeys("a", "b", "gamma");
        rule3.setExactMatchValues("11", "21", "41");
        machine.deletePatternRule(rule3.name, rule3.fields);

        List<String> actual3 = machine.rulesForJSONEvent(e.toString());
[...]

PR #80, opened 2023-03-22
Making deletePatternRule a public method.

baldawar:
We already have `addPatternRule` as public but offer no equivalent for deleting patterns. The method is tested as part of our MachineTests.java so assuming we missed making this public in past (probably because nobody asked until https://github.com/aws/event-ruler/issues/64).

Benchmarks were not necessary for this change. 

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,10 +593,11 @@ public static Range between(final double bottom, final boolean openBottom, final
 ```
 
 Once you have constructed appropriate `Patterns` matchers with these methods, you can use the
-following form of `Machine.addRule()` to add them to your machine:
+following methods to add to or delete from your machine:
 
 ```java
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -593,10 +593,11 @@ public static Range between(final double bottom, final boolean openBottom, final
 ```
 
 Once you have constructed appropriate `Patterns` matchers with these methods, you can use the
-following form of `Machine.addRule()` to add them to your machine:
+following methods to add to or delete from your machine:
```
timbray, line 596: Maybe add a line saying that the cautions expressed in the description of `deleteRule` apply here, i.e. the patterns in the `deletePatternRule` call aren't required to match those in the corresponding `add`.

timbray (approved):

PR #95, opened 2023-05-09
Optimize CI workflow

embano1:
### Description of changes:

- Replaces `adopt` with `temurin` due to [1]
- Use `matrix` strategy testing to simplify the workflow file and allow for extensibility (easily add Java and OS versions)
- Also test against Java 11 (throws warnings due to `1.8` setting in `pom.xml` - nothing to worry, unless we want to change the configuration to avoid those)
- Log maven errors with `--errors`
- Uses caching for dependencies to speed up decrease execution times
- Uses `concurrency` setting to cancel running workflows when new commits are pushed to the PR (reduce usage/resource waste)
- Explicitly define permissions for the workflow to reduce attack surface (`{}` i.e. none)

[1]

> NOTE: AdoptOpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt and adopt-openj9, to temurin and semeru respectively, to keep receiving software and security updates. See more details in the [Good-bye AdoptOpenJDK post](https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/).

Source: https://github.com/actions/setup-java

---

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,4 +1,5 @@
 name: Java Build
+permissions: {}
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -7,16 +8,24 @@ on:
...
+  cancel-in-progress: true
+
 jobs:
   build:
-    runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        os: [ ubuntu-20.04 ]
+        java: [ 8, 11 ]
```
baldawar, line 20: if its easy to add 17 as well, would love to start running builds against it.

embano1: Sure, just take a look at the output of `11` though. IMHO there's maven warnings because the pom hardcodes to `1.8` which produces some maven warnings. I'll amend with `17` and you tell me whether that's good.

baldawar: yup. those are mostly javadoc errors which have a simple fix based on https://github.com/assertj/assertj/issues/1403. I'll look into fixing it tomorrow (unless you beat me to it).

PR #104, opened 2023-06-06
Performance improvements on the addPattern code path. Avoiding lots of unnecessary Set creations.

jonessha:
### Description of changes:

Main strategy was turning SingleByteTransition into an Iterable of itself, and then changing many methods that returned Sets to return Iterables, so that in the n=1 case, we can return a SingleByteTransition instead of a single-element set and avoid the set creation. Also, revised updateTransition in ByteMap to work on the same TreeMap instead of creating a replacement.

In my testing, this gives up to a 25% reduction in GC activity, and a 50% reduction in machine creation latency at p99.9.

---


baldawar:
is the performance high enough to update the number within the readme ?

jonessha:
> is the performance high enough to update the number within the readme ?
The performance improvement is just during machine creation, but not during matching. I don't think we have machine creation latency numbers in the README?

jonessha:
> oh you're right. Odd that we don't have a benchmark for this. Feels like a miss. Worth an issue?

https://github.com/aws/event-ruler/issues/105

baldawar (approved):

Issue #44, opened 2022-09-13
Make Benchmark Tests Run as part of CI for Pull-Requests


baldawar:
## What is your idea?

Right now its kind of annoying to ask for benchmark results as part of each pull-requests. Ideally, we'd have benchmarks run as part of the PR or on-command whenever someone wants it before merging a change. 

This would have made the optimization PRs like https://github.com/aws/event-ruler/pull/37 and https://github.com/aws/event-ruler/pull/39 easier to merge in. Also, it reduces one more friction point for any future pull-requests.  

## Would you be willing to make the change? 

Maybe

## Additional Context

When we get this working, we should make sure benchmarks are only ever run as PR and not part of regular builds (where something running for 6~7 mins is painfully slow for development cycles).

embano1:
I can help with the GHA part of this (including a small tweak to cancel long-running jobs such as benchmarks when new commits are pushed to a PR. Since I'm not an expert in Java benchmarking and the tooling, e.g. which output to use, creating +/- comparisons, etc. it would be great if someone could point me in the right direction so I can create the PR.

timbray:
Hmm, I agree in principle but I think people shouldn't send us PRs that have big performance penalties because that will never be OK for Ruler. So I'd like to have some mechanism to provide an incentive to make this clear to a contributor. Maybe it's enough to write it into our submission guidelines or template?

baldawar:
The default template does ask for benchmarks which we can keep as it. The action here should help ensure as folks push additional updates to their PR, the github actions are good enough. 

On steps to implement this, I suspect it'll be like 
* use maven surefire plugin to run a specific test https://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html
  * This is the java class that needs to be run https://github.com/aws/event-ruler/blob/main/src/test/software/amazon/event/ruler/Benchmarks.java
* update https://github.com/aws/event-ruler/blob/main/.github/workflows/CI.yml to setup automation
  * if there's any dependency changes required it'll be within https://github.com/aws/event-ruler/blob/main/pom.xml with this new configuration

baldawar:
closed by https://github.com/aws/event-ruler/pull/178

PR #109, opened 2023-08-15
Port newer benchmarks to JMH and run those benchmarks using GitHub Actions

Liversticks:
### Issue #, if available:

#44 

### Description of changes:

Port the newer benchmarks (the ones using `citylots2.json` to Java Microbenchmark Harness (JMH). These benchmarks measure throughput in operations/second. The project has been restructured so the JMH benchmarks live in the `benchmark` module and the main library code lives in the `ruler` module.

Existing benchmarks are kept for correctness check (they have JUnit asserts).

Add a new step in CI that runs the benchmarks after running the library tests. We can change the number of iterations (warmup/regular) which affects how much time the benchmarks take to run.

By default, JMH runs 5 warmup iterations and 5 regular iterations. This current benchmark suite takes about 2 hours and 20 minutes to run.

---


baldawar:
Hey Oliver,
thanks for getting out so quick. I will take a deeper look tomorrow evening but had two questions upfront 

1/ is the pkg restructure necessary? If so, would it be better to it in a separate PR to avoid coupling with you change (or is that a lot of hassle)
2/ does the benchmark needs its own pom file if we add JMH as a dependency in the main ruler pkg ? I'm worried that with two POM files we might eventually diverge the dependency chain.

ruler/src/test/software/amazon/event/ruler/Benchmarks.java
baldawar, line 1: I guess this class is no longer necessary

baldawar:
test themselves LGTM but the folder restrucutre still bothers me. the other 2 hr runtime is concerining  as well. Do you know if that's happening because of we build against multiple version of java (7, 11, and 17).

dummy.md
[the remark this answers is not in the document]
Liversticks: It's not, I added it to trigger a commit for CI. I'll remove it.

ruler/src/test/software/amazon/event/ruler/Benchmarks.java
Liversticks: Makes sense, the methods under benchmark are still tested in the other tests so I will remove the Benchmarks class.

Liversticks:
> test themselves LGTM but the folder restrucutre still bothers me. the other 2 hr runtime is concerining as well. Do you know if that's happening because of we build against multiple version of java (7, 11, and 17).

By default, JMH runs 5 warmup iterations and 5 iterations for each benchmark and runs each benchmark in 5 forks. Each iteration (warmup and regular) takes 10 seconds. There are 17 benchmarks. Not including overhead time (example: reading in citylots2 file), the default time is hence `17 * (5 * (10 *(5 + 5))) = 8500 s` or 2.36 hours. The CI builds against multiple versions of Java in parallel so that is not the root cause of the long runtime.

To reduce runtime, we can lower the parameters. If we run 1 warmup iteration, 3 regular iterations, 3 forks, and set the iteration runtime to 5 s, the new runtime is `17 * (3 * (5 * (3 + 1))) = 1020 s` or 17 minutes.

baldawar:
> To reduce runtime, we can lower the parameters. If we run 1 warmup iteration, 3 regular iterations, 3 forks, and set the iteration runtime to 5 s, the new runtime is 17 * (3 * (5 * (3 + 1))) = 1020 s or 17 minutes. 

17 mins is fine IMO. Ideally, we'd want to run the shorter runs of every PR and then periodically run the longer tests (like once a week) but that can be done later.

Liversticks:
> Hey Oliver, thanks for getting out so quick. I will take a deeper look tomorrow evening but had two questions upfront
> 
> 1/ is the pkg restructure necessary? If so, would it be better to it in a separate PR to avoid coupling with you change (or is that a lot of hassle) 2/ does the benchmark needs its own pom file if we add JMH as a dependency in the main ruler pkg ? I'm worried that with two POM files we might eventually diverge the dependency chain.

I'd like to avoid the package restructure but it's the recommended way to add JMH in: https://github.com/openjdk/jmh. From the README there:
> NOTE: JMH is not intended to be used in the same way as a typical testing library such as JUnit. Simply adding the jmh-core jar file to your build is not enough to be able to run benchmarks.

PR #110, opened 2023-08-23
Custom html rule

ajithKeerikkattil:


baldawar:
hey @KQMR-tista I know this is WIP for you but would be keen to understand the use-case so that I can better assist you.

timbray (commented):
There is no documentation, i.e. no .md file updates, so I can't figure out what this is for.

PR #114, opened 2023-10-06
Bringing Ruler packages back in sync

baldawar:
### Description of changes:

Both packages had bifurcated when wildcard and ruler internal improvements were happening in parallel. Aligning code-bases before move everyone to the OSS version under the cover. 

---

### DIFF
--- a/src/main/software/amazon/event/ruler/ByteMachine.java
+++ b/src/main/software/amazon/event/ruler/ByteMachine.java
@@ -91,7 +91,7 @@ class ByteMachine {
         boolean fieldValueIsNumeric = false;
         if (hasNumeric.get() > 0) {
             try {
-                final double numerically = JavaDoubleParser.parseDouble(valString);
+                final double numerically = Double.parseDouble(valString);
                 valString = ComparableNumber.generate(numerically);
                 fieldValueIsNumeric = true;
             } catch (Exception e) {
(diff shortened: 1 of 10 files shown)


src/main/software/amazon/event/ruler/ByteMachine.java
```diff
@@ -91,7 +90,7 @@ Set<NameStateWithPattern> transitionOn(String valString) {
         boolean fieldValueIsNumeric = false;
         if (hasNumeric.get() > 0) {
             try {
-                final double numerically = JavaDoubleParser.parseDouble(valString);
```
rudygt, line 94: can we keep our version here, I am pretty sure this one is faster than the standar double parser

baldawar: Hey Rudy, 
that's a good catch. let me double check the benchmarks. I thought JavaDoubleParser was faster too. I'll will respond on Tuesday (away until then).

baldawar: I don't see any regressions or improvements within the benchmarks because most of the time is spent in `ComparableNumber` but when testing in isolation against the numbers within the benchmark data, I see this FastDoubleParser to be slightly better (45 seconds vs 50 seconds). Added it back for now.

jonessha (approved):

PR #131, opened 2023-11-22
Cidr bug fixes

jonessha:
### Description of changes:

There were two bugs with CIDR:

1. When an IP pattern was present and the incoming value was also an IP, we would convert the IP to a hex string before performing matching. Problem with this was that if there was also a rule present that used anything-but on this same field, the anything-but would always be satisfied, since the anything-but does not specify our internal hex format.
2. When both a CIDR pattern and a numeric pattern were present and the incoming value was an IP, we would attempt to convert to a numeric value first, which would fail, and cause us to skip over CIDR matching and go straight to String matching.

```
EXACT events/sec: 169505.2
WILDCARD events/sec: 127968.8
PREFIX events/sec: 166981.2
PREFIX_EQUALS_IGNORE_CASE_RULES events/sec: 177113.9
SUFFIX events/sec: 190579.6
SUFFIX_EQUALS_IGNORE_CASE_RULES events/sec: 194051.0
EQUALS_IGNORE_CASE events/sec: 174075.2
NUMERIC events/sec: 115359.0
ANYTHING-BUT events/sec: 126524.9
ANYTHING-BUT-IGNORE-CASE events/sec: 115109.7
ANYTHING-BUT-PREFIX events/sec: 128509.0
ANYTHING-BUT-SUFFIX events/sec: 130716.6
COMPLEX_ARRAYS events/sec: 30098.6
PARTIAL_COMBO events/sec: 51279.9
COMBO events/sec: 19901.7
```

---

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
   <groupId>software.amazon.event.ruler</groupId>
   <artifactId>event-ruler</artifactId>
   <name>Event Ruler</name>
-  <version>1.6.0</version>
+  <version>1.7.0</version>
   <description>Event Ruler is a Java library that allows matching Rules to Events. An event is a list of fields,
     which may be given as name/value pairs or as a JSON object. A rule associates event field names with lists of
     possible values. There are two reasons to use Ruler: 1/ It's fast; the time it takes to match Events doesn't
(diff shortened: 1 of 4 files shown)


timbray:
Hey, those are *my* bugs you're killing.

baldawar (approved):

PR #143, opened 2024-02-23
Supporting lists of values for anything-but-prefix and anything-but-suffix

jonessha:
### Description of changes:
Supporting lists of values for anything-but-prefix and anything-but-suffix. Just like for anything-but-equals-ignore-case.

---


src/main/software/amazon/event/ruler/JsonRuleCompiler.java
```diff
@@ -608,25 +594,41 @@ private static Patterns processAnythingButListMatchExpression(JsonParser parser)
...
                 switch (token) {
                     case VALUE_STRING:
-                        values.add('"' + parser.getText() + '"');
+                        String text = parser.getText();
+                        if (matchType != MatchType.ANYTHING_BUT_IGNORE_CASE && text.isEmpty()) {
+                            barf(parser, "Null prefix/suffix not allowed");
+                        }
+                        values.add((matchType != MatchType.ANYTHING_BUT_SUFFIX ? "\"" : "") +
+                                   text +
+                                   (matchType != MatchType.ANYTHING_BUT_PREFIX ? "\"" : ""));
```
baldawar, line 611: thinking long-term, this is an easy place to introduce new bugs with multiple conditions in the same place. Possible to extract this into a method?

src/main/software/amazon/event/ruler/JsonRuleCompiler.java
```diff
@@ -608,25 +594,41 @@ private static Patterns processAnythingButListMatchExpression(JsonParser parser)
...
                         break;
                     default:
-                        barf(parser, "Inside anything-but/equals-ignore-case list, number|start|null|boolean is not supported.");
+                        if (matchType == MatchType.ANYTHING_BUT_IGNORE_CASE) {
+                            barf(parser, "Inside anything-but/equals-ignore-case list, number|start|null|boolean is not supported.");
```
baldawar, line 615: can be more helpful for debugging if we change this to `"Inside anything-but/equals-ignore-case, " + token +" is not supported"`

src/main/software/amazon/event/ruler/Patterns.java
```diff
@@ -85,12 +86,22 @@ public static AnythingBut anythingButNumberMatch(final Set<Double> anythingButs)
...
+    }
+
+    public static AnythingButValuesSet anythingButSuffix(final String suffix) {
+        return new AnythingButValuesSet(MatchType.ANYTHING_BUT_SUFFIX,
+                Collections.singleton(new StringBuilder(suffix).reverse().toString()));
+    }
+
+    public static AnythingButValuesSet anythingButSuffix(final Set<String> anythingButs) {
+        return new AnythingButValuesSet(MatchType.ANYTHING_BUT_SUFFIX,
+                anythingButs.stream().map(s -> new StringBuilder(s).reverse().toString()).collect(Collectors.toSet()));
```
baldawar, line 104: Do these need to be mentioned in README ? in the patterns API section?

src/main/software/amazon/event/ruler/JsonRuleCompiler.java
jonessha: Perhaps, but this would require changing outward-facing error messages for existing types, which I'd prefer not to do without a strong reason. Also, one benefit to the existing message is it says _why_ the token is not supported. If we just print the token there, it might not be obvious it is of the wrong type.

src/main/software/amazon/event/ruler/Patterns.java
jonessha: That section was quite out of date. I've brought it up to date.

baldawar (approved):

Issue #72, opened 2023-02-11
Support for anything-but/wildcard


schenksj:
## What is your idea?
Support wildcard matches in anything-but:

```
{
  "detail": {
    "command_path": [ { "anything-but": {"wildcard": "*cmd.exe*mycommand*"} } ]
  }
}
```

## Would you be willing to make the change? 
Probably

## Additional context

Add any other context (such as images, docs, posts) about the idea here.

baldawar:
Gotta be honest, was expect this to pop up eventually. 

I don't have objections to the feature itself but there are performance implications of using wildcards, so benchmarks need to be thorough. Wildcards have also not been validated against production traffic (or at least nobody has shared it with me), so I'd encourage additional testing within the library but also within your end application.

We also have a new effort with this class that helps reason complexity tied to wildcards https://github.com/aws/event-ruler/blob/main/src/main/software/amazon/event/ruler/MachineComplexityEvaluator.java, but it won't work because of line 178.

schenksj:
Thanks @baldawar !  I'll probably take a look at this at some point after #71 is done, its not terribly high priority for me but it only looks like a few minutes of work.  Quick question, is my assumption that wildcard matching only accepts one pattern (not a list) as input correct?

baldawar:
I'm expecting this has some trickiness. Wildcard involves compound byte transitions and opens risk of forever looping. The MachineComplextyEvaluator is also an unknown in if it'll work. 

In general, wildcard performance can downgrade performance (see https://github.com/aws/event-ruler#matching-with-a-machine for some numbers), so I'm wondering if it'll be worth the effort.

schenksj:
@baldawar I think for now anything-but enhancements are lower priority for me.  I came to terms with the limitations and in my app have a secondary "if-matches-exclude" rule capability that can be applied on top of rules.  The bigger annoyance for now is around case-insensitivity in wildcards (and suffix/prefix though you can accomplish similar work with wildcards).  My work-around is less elegant (pre-processing to re-present fields in lowercase before running matches).

I'm comfortable shelving this one for now.  Thank you for your thoughts.

baldawar:
I'm going to leave the issues open for a while in case it someone else wants to dig into them later.  

on 

>  The bigger annoyance for now is around case-insensitivity in wildcards (and suffix/prefix though you can accomplish similar work with wildcards). My work-around is less elegant (pre-processing to re-present fields in lowercase before running matches).

[...]

PR #145, opened 2024-03-12
Anything but wildcard

jonessha:
### Description of changes:
Adding support for anything-but-wildcard.

Issue: https://github.com/aws/event-ruler/issues/72

On my machine, performance is similar to all of the other anything-buts. Can get an exact number from "Pull request checks -> Java build -> build (ubuntu-X.Y, 8) -> Run benchmarks".

---

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
   <groupId>software.amazon.event.ruler</groupId>
   <artifactId>event-ruler</artifactId>
   <name>Event Ruler</name>
-  <version>1.7.3</version>
+  <version>1.7.4</version>
   <description>Event Ruler is a Java library that allows matching Rules to Events. An event is a list of fields,
     which may be given as name/value pairs or as a JSON object. A rule associates event field names with lists of
     possible values. There are two reasons to use Ruler: 1/ It's fast; the time it takes to match Events doesn't
--- a/src/main/software/amazon/event/ruler/ByteMachine.java
+++ b/src/main/software/amazon/event/ruler/ByteMachine.java
@@ -149,6 +149,7 @@ class ByteMachine {
             case ANYTHING_BUT_IGNORE_CASE:
             case ANYTHING_BUT_PREFIX:
             case ANYTHING_BUT_SUFFIX:
+            case ANYTHING_BUT_WILDCARD:
                 assert pattern instanceof AnythingButValuesSet;
                 deleteAnythingButValuesSetPattern((AnythingButValuesSet) pattern);
                 break;
@@ -532,6 +533,7 @@ class ByteMachine {
                             }
                             break;
                         case ANYTHING_BUT_IGNORE_CASE:
+                        case ANYTHING_BUT_WILDCARD:
                             // only applies if at last character
                             if (valIndex == (val.length - 1)) {
                                 addToAnythingButsMap(failedAnythingButs, match.getNextNameState(), match.getPattern());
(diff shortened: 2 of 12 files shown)


src/main/software/amazon/event/ruler/ByteMachine.java
```diff
@@ -1680,16 +1684,15 @@ private void addMatchReferences(ByteMatch match) {
                 hasNumeric.incrementAndGet();
             }
             break;
+        case ANYTHING_BUT_PREFIX:
         case ANYTHING_BUT_IGNORE_CASE:
+        case ANYTHING_BUT_WILDCARD:
             addToAnythingButsMap(anythingButs, match.getNextNameState(), match.getPattern());
             break;
         case ANYTHING_BUT_SUFFIX:
             hasSuffix.incrementAndGet();
```
baldawar, line 1693: A lot of the special logic around hasNumeric / hasSuffix can be abstracted away if there was method here that recursively called the same switch function. It an overkill to be combined with this change, but a follow-up change would help make anything-but way more compatible with all future matchers (and remove the ANYTHING_BUT_<MATCHER> constants.

src/test/software/amazon/event/ruler/ACMachineTest.java
```diff
@@ -2564,4 +2757,16 @@ public void testAdditionalNameStateReuseSecondSubRuleSubsetOfFirstSubRule() thro
         assertEquals(1, matches.size());
         assertTrue(matches.contains("rule1"));
     }
+
+    private static Set<String> set(String ... strings) {
```
baldawar, line 2761: nit: accidental space between String and ...

src/test/software/amazon/event/ruler/Benchmarks.java
```diff
@@ -746,6 +775,12 @@ public void CL2Benchmark() throws Exception {
 
         bm = new Benchmarker();
 
+        bm.addRules(ANYTHING_BUT_WILDCARD_RULES, ANYTHING_BUT_WILDCARD_MATCHES);
+        bm.run(citylots2);
+        System.out.println("ANYTHING-BUT-WILDCARD events/sec: " + String.format("%.1f", bm.getEPS()));
```
baldawar, line 780: anything worth updating on the readme ?

pom.xml
```diff
@@ -20,7 +20,7 @@
   <groupId>software.amazon.event.ruler</groupId>
   <artifactId>event-ruler</artifactId>
   <name>Event Ruler</name>
-  <version>1.7.3</version>
+  <version>1.7.4</version>
```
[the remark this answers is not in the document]
jonessha: Ah, right. Changed it back.

src/test/software/amazon/event/ruler/ACMachineTest.java
jonessha: Oh really? This was intentional. It's the way I've always done it. Is it generally considered better form to omit the space?

src/test/software/amazon/event/ruler/Benchmarks.java
jonessha: Nah cause README mentions top-level anything-but, but doesn't go into the subtypes. Which I think is fine, because they all seem to have the same performance.

src/test/software/amazon/event/ruler/ACMachineTest.java
baldawar: don't think one is better than the other, but I thought we were using no space in the pkg. You comment made me check and realize I had it the wrong way. we do use space everywhere https://github.com/search?q=repo%3Aaws%2Fevent-ruler+...+path%3A%2F%5Esrc%5C%2F%2F&type=code, so ignore my previous comment.

baldawar (approved):

PR #173, opened 2024-08-09
Optimization: removed level of indirection for SubRuleContext.

svladykin:
### Description of changes:

Instead of `Double` id use `SubRuleContext` directly everywhere. `SubRuleContext` id is only used internally for backward compatibility in `equals` and `hashCode` (not sure if this is actually needed, but tests check that). The benchmarks are not stable (need JMH with forking, proper warmup and multiple longer iterations for individual benchmarks), but overall can see 3-8% in throughput improvement on some benchmarks with this easy fix.

Using `long` as id instead of `double` because `long` is typically faster and actually can produce more unique 64-bit patterns than `double` because for `long` every 64-bit pattern is valid while for `double` it is not true (NaN).

Changed only types but did not rename variables all over the place because it would be harder to review, can be done later as needed.

```Reading citylots2
Read 213068 events
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
Lines: 213068, Msec: 11459
Events/sec: 18593.9
 Rules/sec: 130157.6
Reading citylots2
Read 213068 events
EXACT events/sec: 274218.8
WILDCARD events/sec: 176526.9
PREFIX events/sec: 271078.9
PREFIX_EQUALS_IGNORE_CASE_RULES events/sec: 272465.5
SUFFIX events/sec: 260474.3
SUFFIX_EQUALS_IGNORE_CASE_RULES events/sec: 266002.5
EQUALS_IGNORE_CASE events/sec: 233883.6
NUMERIC events/sec: 145937.0
ANYTHING-BUT events/sec: 129919.5
ANYTHING-BUT-IGNORE-CASE events/sec: 142902.7
ANYTHING-BUT-PREFIX events/sec: 153507.2
ANYTHING-BUT-SUFFIX events/sec: 150365.6
ANYTHING-BUT-WILDCARD events/sec: 160322.0
COMPLEX_ARRAYS events/sec: 32078.9
PARTIAL_COMBO events/sec: 49037.5
COMBO events/sec: 20434.3
```

Overall the becnhmark results are not conclusive across diffrent jvm versions: the same benchmark can be noticeably better or worse, from what I see from main branch history they are quite flaky. Will try to improve benhcmarks in a separate PR. 
---

### DIFF
--- a/src/main/software/amazon/event/ruler/NameState.java
+++ b/src/main/software/amazon/event/ruler/NameState.java
@@ -36,13 +36,13 @@ class NameState {
     private final Map<Patterns, Set<Object>> patternToRules = new ConcurrentHashMap<>();
 
     // All terminal sub-rule IDs, keyed by pattern, that led to this NameState.
-    private final Map<Patterns, Set<Double>> patternToTerminalSubRuleIds = new ConcurrentHashMap<>();
+    private final Map<Patterns, Set<SubRuleContext>> patternToTerminalSubRuleIds = new ConcurrentHashMap<>();
 
     // All non-terminal sub-rule IDs, keyed by pattern, that led to this NameState.
-    private final Map<Patterns, Set<Double>> patternToNonTerminalSubRuleIds = new ConcurrentHashMap<>();
+    private final Map<Patterns, Set<SubRuleContext>> patternToNonTerminalSubRuleIds = new ConcurrentHashMap<>();
 
     // All sub-rule IDs mapped to the number of times that sub-rule has been added to this NameState.
-    private final Map<Double, Integer> subRuleIdToCount = new ConcurrentHashMap<>();
+    private final Map<SubRuleContext, Integer> subRuleIdToCount = new ConcurrentHashMap<>();
 
     ByteMachine getTransitionOn(final String token) {
         return valueTransitions.get(token);
@@ -90,7 +90,7 @@ class NameState {
      * @param pattern The pattern that the rules must use to get to this NameState.
      * @return The sub-rules, could be null if none for pattern.
      */
-    Set<Double> getTerminalSubRuleIdsForPattern(Patterns pattern) {
+    Set<SubRuleContext> getTerminalSubRuleIdsForPattern(Patterns pattern) {
         return patternToTerminalSubRuleIds.get(pattern);
     }
 
@@ -105,7 +105,7 @@ class NameState {
      * @param pattern The pattern that the rules must use to get to this NameState.
      * @return The sub-rule IDs, could be null if none for pattern.
      */
-    Set<Double> getNonTerminalSubRuleIdsForPattern(Patterns pattern) {
+    Set<SubRuleContext> getNonTerminalSubRuleIdsForPattern(Patterns pattern) {
         return patternToNonTerminalSubRuleIds.get(pattern);
     }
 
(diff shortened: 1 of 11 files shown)


baldawar:
hey @svladykin  is this ready to review? its' been in draft state which has me confused on the state of the PR.

src/main/software/amazon/event/ruler/SubRuleContext.java
```diff
@@ -11,63 +11,52 @@
...
      * Generator for SubRuleContexts.
      */
     static final class Generator {
 
-        private double nextId = -Double.MAX_VALUE;
-
-        private final Map<Object, Set<Double>> nameToIds = new ConcurrentHashMap<>();
-        private final Map<Double, Object> idToName = new ConcurrentHashMap<>();
+        private final Map<Object, Set<SubRuleContext>> nameToContext = new ConcurrentHashMap<>();
+        private long nextId;
```
baldawar, line 50: Should this be AtomicLong ?

src/main/software/amazon/event/ruler/NameState.java
```diff
@@ -175,12 +175,11 @@ boolean isEmpty() {
      * @param pattern The pattern used by the sub-rule to transition to this NameState.
      * @param isTerminal True indicates that the sub-rule is using pattern to match on the final event field.
      */
-    void addSubRule(final Object rule, final double subRuleId, final Patterns pattern, final boolean isTerminal) {
+    void addSubRule(final Object rule, final SubRuleContext subRuleId, final Patterns pattern, final boolean isTerminal) {
```
baldawar, line 178: wondering if here and elsewhere in the class if `rule` can be fetched from `subRuleId` instead

src/main/software/amazon/event/ruler/SubRuleContext.java
svladykin: From what I see the only place where it is called is `GenericMachine.addPatternRule()` under `synchronized(this)`, so `AtomicLong` does not make much sense.

src/main/software/amazon/event/ruler/NameState.java
svladykin: This is possible, but looks like if we do that everywhere the patch will be much larger, at the same time I don't see much improvement from this change. Either option is fine with me.

svladykin:
@baldawar I did not like the `Benchmark` numbers for this PR, thus submitted a second one with JMH benchmarks, assuming that I will be able to get more meaningful perfromance numbers here when the second PR is merged. This is why it is still in draft status.

src/main/software/amazon/event/ruler/NameState.java
baldawar: I think that's an acceptable next step for this PR, though digging into JMH variances would be better place to focus right now.

svladykin:
Rebased on top of main to check JMH throughput numbers.

baldawar:
odd, perf tests are a bit of mixed bag when compared to https://github.com/aws/event-ruler/actions/runs/10805542552/job/30018080464 . Is it because of most of the time is spent in other parts of the code (my recent tests showed `ByteMachine.getTransitionOn` and code tied to `exists` matcher to be quite expensive)

svladykin:
Yes, I don't see any throughput improvement as well. We can either consider this patch to be just a code cleanup or just close it and stop spending time on it. I'm good with either option.

I also noticed that json parsing takes around 60% of the time for simple rules, which makes me think that using custom binary format instead of json could help a lot.

timbray:
FWIW, Quamina has a custom hand crafted JSON parser for events, and the benefits of that were huge. But didn't bother for rule parsing.

baldawar:
useful bookmark for the custom JSON parser : 
* https://www.tbray.org/ongoing/When/202x/2022/06/10/Quamina-Optimizing
* https://github.com/timbray/quamina/blob/main/flatten_json.go

For this change, I think its worth merging if we can add a test that shows SubRuleContext is now faster. Benchmark would be over kill but unit test would be good enough with bunch of for-loops similar to `ComparableNumberTest.java` is good enough IMO.

baldawar:
@svladykin let me know if you're still working on this.

svladykin:
Yes, was a bit busy lately, will catch up this week.

svladykin:
Added a simple performance test which takes ~2500ms on the old code and ~1400ms on the new code.

baldawar (approved):

baldawar:
Thanks @svladykin

Issue #174, opened 2024-08-15
Cannot create GenericMachine using builder() for anything but Object without compiler warnings/errors


youngm:
I'd like to create a GenericMachine for Integer and I cannot do it without unsafe casts and compiler warnings.

## To Reproduce
Try these various ways to create a GenericMachine<Integer> without a warning or error. They all fail.
```
GenericMachine<Integer> try1Deprecated = new GenericMachine<Integer>();
GenericMachine<Integer> try1Raw = GenericMachine.builder().build();
GenericMachine<Integer> try2Raw =
    GenericMachine.builder().<GenericMachine<Integer>>build();
GenericMachine<Integer> try3NotVisible =
    new GenericMachine.Builder<GenericMachine<Integer>>().build();

```

## Expected behavior
Should be able to create a GenericMachine for a given type without a compiler warning.

Perhaps the build() method could accept a generic parameter?

## Environment:
Java 17

## Additional context
Sure, GenericMachine<Object> works but why make it generic if there isn't a clean way to make an instance of any other type?

baldawar:
Merged https://github.com/aws/event-ruler/pull/175. thanks @youngm

PR #175, opened 2024-08-16
Improve GenericMachine.Builder Generics

youngm:
### Issue #, if available:
[174](https://github.com/aws/event-ruler/issues/174)

### Description of changes:
I Improved the Builder Generics to allow GenericMachine.Builder to create instances of a type while still allowing Machine.Builder to function.

```
GenericMachine<Integer> machine = GenericMachine.<Integer>builder().build();
```

### DIFF
--- a/src/main/software/amazon/event/ruler/GenericMachine.java
+++ b/src/main/software/amazon/event/ruler/GenericMachine.java
@@ -1,9 +1,9 @@
 package software.amazon.event.ruler;
 
-import static software.amazon.event.ruler.SetOperations.intersection;
-
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonNode;
+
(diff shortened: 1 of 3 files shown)


baldawar:
@youngm thanks for catching the bug and the PR. I think the fix makes sense. Could you also add a unit test to make sure we don't break type by error.

youngm:
@baldawar I wasn't sure how to test this since I was just trying to avoid a compiler warning. So, I just added a test that doesn't use a String as the type. Does that work?

timbray (approved):
Just wanted to say that it's cool that this works.

baldawar:
>  I just added a test that doesn't use a String as the type. Does that work?

Yup. Good enough for me. Thanks. I'll merge this once the builds pass.

baldawar (approved):

Issue #179, opened 2024-08-28
Improve numeric matching


timbray:
## What is your idea?

For Quamina, a couple of folks figured how to represent the whole range of 64-bit float values in 64 big-endian bits, and then to encode them in base128, then to discard certain suffixes.  Check out numbits.go https://github.com/timbray/quamina/blob/main/numbits.go

So you get a smaller size representation of numeric field values, and no more subsetting of the the numbers that can be matched.

I can't see any reason this wouldn't work in Ruler .

## Would you be willing to make the change? 

I seem to recall that Rishi just wired in a similar flavor of change, so I suggest this one would be easy for him.

baldawar:
oh that's neat! 

This should be doable within ruler though I won't be able to pick up for few weeks due to an internal launch. 

Placing down links of the files I'd expect we'd need to touch to enforce this https://github.com/aws/event-ruler/commit/d04e3f03bd68d583cd7e01d9e174ad8ff949ac7b#diff-58bfacbbf2f6f6e26165ed131f0cae9667cf41d642d57b1a154ca97236507bce. 

To keep codebases roughly similar, I'll try to imitate numbits.go as much as Java lets me.

timbray:
BTW I wrote a blog about it at https://www.tbray.org/ongoing/When/202x/2024/08/28/Q-Numbers-2 and Arne Hoffman has promised to write an explanation of bit-masking voodoo, will put a pointer in here when I see it.  Off the top of my head, I don't think Java should get in the way, although I'm not sure there's an equivalent of Go's `math.Float64bits(f)`.  It’s a little weird because the byte values are between 0 and 127 inclusive, a lot of which are *not* printable characters even though they are valid UTF-8.  In Quamina we do a little extra work to shorten the 10-byte results where possible but I think that's going to screw up the horrible Range model-building logic; shouldn't make it impossible but it will have to be modified. If it were me I might decide to leave it at 10 bytes just to avoid that work.

baldawar:
Thanks Tim. Cursory check points me to [Double.doubleToLongBits](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#doubleToLongBits-double-) and related functions. I haven't had the time to explore how these methods behaves across various scenarios but hopefully it good enough for ruler's needs.

PR #188, opened 2024-09-18
Improve Numeric matching to support full range of float64

baldawar:
### Issue #, if available: 

179

### Description of changes: 

This change follows the guidance from https://github.com/aws/event-ruler/issues/179 on using 10 byte base-128 encoded format for numbers similar to how Quamina does it. 

Didn't see any performance implications of supporting the new range, but had to fix a bunch of tests. I will be changing the numbers we use for testing to better test the new range of numbers before merging. 

During debugging, I found it challenging to make sense of the numbers to I've also added a helper method in `ComparableNumbers` and modified `toString()` methods in few places. 

```
Benchmark                                                  (benchmark)   Mode  Cnt       Score       Error  Units
CityLots2JmhBenchmarks.group01Simple                             EXACT  thrpt   10  413930.986 ± 11450.753  ops/s
CityLots2JmhBenchmarks.group01Simple                          WILDCARD  thrpt   10  352480.174 ±  6750.778  ops/s
CityLots2JmhBenchmarks.group01Simple                            PREFIX  thrpt   10  418264.642 ± 19729.068  ops/s
CityLots2JmhBenchmarks.group01Simple   PREFIX_EQUALS_IGNORE_CASE_RULES  thrpt   10  415142.905 ± 18319.570  ops/s
CityLots2JmhBenchmarks.group01Simple                            SUFFIX  thrpt   10  419590.039 ± 15463.837  ops/s
CityLots2JmhBenchmarks.group01Simple   SUFFIX_EQUALS_IGNORE_CASE_RULES  thrpt   10  410364.141 ± 12610.649  ops/s
CityLots2JmhBenchmarks.group01Simple                EQUALS_IGNORE_CASE  thrpt   10  382570.869 ±  5316.951  ops/s
CityLots2JmhBenchmarks.group01Simple                           NUMERIC  thrpt   10  256233.479 ±  3662.693  ops/s
CityLots2JmhBenchmarks.group01Simple                      ANYTHING-BUT  thrpt   10  276676.782 ±   958.215  ops/s
CityLots2JmhBenchmarks.group01Simple          ANYTHING-BUT-IGNORE-CASE  thrpt   10  274642.171 ±  5587.937  ops/s
CityLots2JmhBenchmarks.group01Simple               ANYTHING-BUT-PREFIX  thrpt   10  288814.025 ±  1302.395  ops/s
CityLots2JmhBenchmarks.group01Simple               ANYTHING-BUT-SUFFIX  thrpt   10  284559.927 ±  3258.949  ops/s
CityLots2JmhBenchmarks.group01Simple             ANYTHING-BUT-WILDCARD  thrpt   10  303269.268 ±   922.894  ops/s
CityLots2JmhBenchmarks.group02Complex                   COMPLEX_ARRAYS  thrpt    6   54274.020 ±   567.870  ops/s
CityLots2JmhBenchmarks.group02Complex                    PARTIAL_COMBO  thrpt    6  102749.040 ±  2765.021  ops/s
CityLots2JmhBenchmarks.group02Complex                            COMBO  thrpt    6   35555.689 ±   312.623  ops/s
```
---


src/main/software/amazon/event/ruler/ComparableNumber.java
```diff
@@ -3,104 +3,122 @@
...
-        final long shiftedBySixDecimals = number.multiply(TEN_E_SIX).longValueExact();
+        final BigDecimal bigDecimal = JavaBigDecimalParser.parseBigDecimal(str);
+        final double doubleValue = bigDecimal.doubleValue();
 
-        // faster than doing bigDecimal comparisons
-        if (shiftedBySixDecimals < -HALF_TRILLION_TEN_E_SIX || shiftedBySixDecimals > HALF_TRILLION_TEN_E_SIX) {
-            throw new IllegalArgumentException("Value must be between " + -ComparableNumber.HALF_TRILLION +
-                    " and " + ComparableNumber.HALF_TRILLION + ", inclusive");
+        // make sure we have the comparable numbers and haven't eaten up decimals values
+        if(Double.isNaN(doubleValue) || Double.isInfinite(doubleValue) ||
```
timbray, line 54: Can this happen? JSON doesn't allow numeric values of NaN and Inf

src/main/software/amazon/event/ruler/ComparableNumber.java
```diff
@@ -3,104 +3,122 @@
...
+     * The method first determines the number of trailing zero bytes in the input
+     * value by iterating over the bytes from the most significant byte to the least
+     * significant byte, and counting the number of consecutive zero bytes at the end.
+     * It then creates a byte array of fixed length {@code MAX_LENGTH_IN_BYTES} and
+     * populates it with the Base128 encoded bytes of the input value, starting from
+     * the least significant byte.
+     * <br/>
+     * As shown in Quamina's numbits.go, it's possible to use variable length encoding
+     * to reduce storage for simple (often common) numbers but it's not done here to
+     * keep range comparisons simple for now.
```
timbray, line 83: Haha, ever since Quamina got variable-width numbits I've been worrying about the range comparison code.  *sigh* now I guess I have to figure that out.

src/main/software/amazon/event/ruler/ComparableNumber.java
```diff
@@ -3,104 +3,122 @@
...
+     *
+     * @param value the long value to be converted
+     * @return the Base128 encoded string representation of the input value
+     */
+    public static String numbits(long value) {
```
timbray, line 88: Each char in a java String is 2 bytes because they are UTF-16 codepoints. You could save a lot of memory if you could work on the UTF-8 bytes. But I realize that this comment applies to all of Ruler I guess.

timbray (approved):
LGTM, a couple of nonblocking comments to look at

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: not in JSON but there's interest in supporting other formats so I'm bullet proofing this now than be in for a surprise in future.

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: Yeah. I gave it a shot but found it hard to get right. I wanted to get back to building $and support, so took the easier route. I figured I can come back to cleaning up my branch later.

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: that's fair. It would be worth a look when we dig into memory & cpu improvements down the line.

src/main/software/amazon/event/ruler/ComparableNumber.java
```diff
@@ -3,104 +3,122 @@
...
+ * We use Base128 encoding offers a compact representation of decimal numbers
+ * as it preserves the lexicographical order of the numbers. See
+ * https://github.com/aws/event-ruler/issues/179 for more context.
  * <br/>
- * The number is parsed as a Java {@code BigDecimal} to support decimal fractions. We're avoiding double as
- * there is a well-known issue that double numbers can lose precision when performing calculations involving
- * other data types. The higher the number, the lower the accuracy that can be maintained. For example,
- * {@code 0.30d - 0.10d = 0.19999999999999998} instead of {@code 0.2d}. When extended to {@code 1e10}, the test
- * results show that only 5 decimal places of precision can be guaranteed when using doubles.
+ * The numbers are first parsed as a Java {@code BigDecimal} as there is a well known issue
```
arnehormann, line 21: Can you provide a reference or details for that?

Double handling is finicky and has lots of potential foot guns. But I never ran into a loss of precision that's not mediated by decimal-binary conversion issues. If you e.g. use Double.toHexString and the "p" representation of exponents in your numerical tests, you should be able to avoid these issues.
This only applies to where you want to use numbers and not the strings, of course. But at some point, the JSON number strings will be converted to double and you'll run into the same issue.
Unless the JSON parsers also avoid Double. But then, you should use the same code path the JSON parser uses.

You won't lose fidelity. But you'll gain some performance.

src/main/software/amazon/event/ruler/ComparableNumber.java
```diff
@@ -3,104 +3,122 @@
...
+                BigDecimal.valueOf(doubleValue).compareTo(bigDecimal) != 0) {
+            throw new IllegalArgumentException("Cannot compare number : " + str);
         }
+        final long bits = Double.doubleToRawLongBits(doubleValue);
 
-        return longToBase64Bytes(shiftedBySixDecimals + HALF_TRILLION_TEN_E_SIX);
+        // if high bit is 0, we want to xor with sign bit 1 << 63, else negate (xor with ^0). Meaning,
+        // bits >= 0, mask = 1000000000000000000000000000000000000000000000000000000000000000
+        // bits < 0,  mask = 1111111111111111111111111111111111111111111111111111111111111111
+        final  long mask = ((bits >>> 63) * 0xFFFFFFFFFFFFFFFFL) | (1L  << 63);
```
arnehormann, line 63: Raph Levien proposed an even nicer way on Mastodon (not yet in Quamina afaik).
`(bits >> 63) | (1L << 63)` is sufficient. For negative numbers, the sign bit will be set. `>>` is sign extending and will fill all the left bits with the sign bit. No need for another constant.

src/main/software/amazon/event/ruler/ComparableNumber.java
arnehormann: How are Strings-by-chars sorted in Event Ruler? Binary value or with some collation? If it's binary... you can still use the same scheme but store 2 7-bit bytes per char. Might require shuffling the bytes around if the endianness of the char 2-byte-tuples is different. But as long as leaving the 0x8080 bits out stays clear of combiners or possibly causing illegal codepoints... you should be good.

src/main/software/amazon/event/ruler/ComparableNumber.java
arnehormann: You can check both in one go based on the bits-representation. With a constant `long NOT_FINITE = 0x7ff0L << 52`: `(bits & NOT_FINITE) == NOT_FINITE` will be true for only NaN and +/- Infinity.

Maybe the JVM automagically already inlines the two calls like that. If not, it could help, here.

arnehormann (commented):
@baldawar/@timbray my 2cts as promised... ping me if you need more details

src/main/software/amazon/event/ruler/ComparableNumber.java
arnehormann: ... or use the lower 15 bits. simpler than a 0x8080 mask. Still, check that you get by with only one codepoint and that it's not reserved or illegal in UTF16

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: `37.807807921694092` was the number I noticed it with.

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: sweet. Thanks. If you have the link to the tweet (toot?) from Raph, I'd love to link it for attribution.

arnehormann: https://mastodon.online/@raph/113071041069390831

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: Comparing ranges got trickier with using 2 7-bit bytes per char, so I've parked the idea for now. Will come back to it later. 

In case anyone takes a stab at it in future, here's the reference code for this : 

```
static final int MAX_LENGTH_IN_BYTES = 5;
...
public static String numbits(long value) {
        char[] result = new char[MAX_LENGTH_IN_BYTES];
        int index = MAX_LENGTH_IN_BYTES - 1;

        while (value != 0 && index >= 0) {
            int lowBits = (int)(value & BASE_128_BITMASK);
            value >>= 7;
            int highBits = (int)(value & BASE_128_BITMASK);
            value >>= 7;

            result[index--] = (char)((highBits << 7) | lowBits);
        }

        return new String(result);
    }
```

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: JVM does inline it. I've left this as it for readability for now.

src/main/software/amazon/event/ruler/ComparableNumber.java
arnehormann: A double represented as a decimal stops being accurate after 15 digits (unless it ends in 5). this number has 17.
A good way to keep it in mind: 10 bits fit 3 decimal digits. a double has 52 mantissa bits, so this buys you a fractional amount more than 15 digits with 50 bits. the other 2 (or 3 if you include the implicit one) are not enough for a 16th digit. So it will... distribute... the rest of the 10 alternatives over 8 states. As powers of two are binary, a number ending in 5 will be at an advantage, here.

timbray: And BTW this is a problem with the citylots file, those floats in there have more precision than is safe with JSON. So it's probably not useful for supporting numeric-match testing.

src/main/software/amazon/event/ruler/ComparableNumber.java
arnehormann: ... Tim told me quamina internally works different. Quamina needs the UTF8 validity because it uses UTF8-illegal byte values as markers. 
Is ruler different enough so you don't depend on that? In that case, you could also just cut it into 4 different shorts. Unless that destroys some string operations you need, e.g by producing hanging combiners etc.

baldawar: For range comparison, Ruler has a lookup table that changes incoming numbers into a sequence of digits https://github.com/aws/event-ruler/blob/main/src/main/software/amazon/event/ruler/Range.java#L128 . Changing this requires more effort and I suspect not as beneficial compared to supporting variable length for numbers. So I chose to work on variable length first and then come back to compressing the numbers later (though I'm hopeful that I'd be able to do both once I get more into refactoring this part of the code more)

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: Oh wow. I did not realize that. Let me look into changing city lots to have a healthy mix of numbers with less than 15 decimals.

src/main/software/amazon/event/ruler/ComparableNumber.java
baldawar: > Quamina needs the UTF8 validity because it uses UTF8-illegal byte values as markers.
> Is ruler different enough so you don't depend on that

I don't know the answer for this on the top of mind. Will check on it later but I was originally suspecting it should be fine as long as I address any edge cases within the look-up table part of the code.

Issue #69, opened 2023-01-30
Support for $and operand


schenksj:
## What is your idea?
It would be great to allow an `$and` operand, (overcoming JSON "last-one-in-wins" semantics for fields) to enable tagging multiple conditions to a given field.  I think this may be possible given how I interpret the implementation of the `$or` operand.  For example, this would exclude any value that (starts with "abc" AND ends with "123" AND also excludes "notmyvalue" and "alsonotmyvalue").

```
{ "$and": [
       {"myField": [{"anything-but": ["notmyvalue", "alsonotmyvalue"]}]},
       {"myField": [{"anything-but": {"suffix": "123"}}]},
       {"myField": [{"anything-but": {"prefix": "abc"}}]}
   ]
}
```

## Would you be willing to make the change? 
Maybe

## Additional context

Add any other context (such as images, docs, posts) about the idea here.

baldawar:
Make sense. I've heard some murmurs on needing this elsewhere as well. Similar to `$or` we'd have to 
* add a new benchmark test (`$or` has some performance concerns, `$and` will too). 
* we'll need to support $and working at different depths and multiple fields

```
{ "topField" : {
    "$and": [
          {"myField": [{"anything-but": ["notmyvalue", "alsonotmyvalue"]}]},
          {"anotherField": [{"anything-but": {"suffix": "123"}}]},
          {"lastField": [{"anything-but": {"prefix": "abc"}}]}
      ]
   } 
}
 ```

schenksj:
@baldawar - I suspect the above example doesn't need the $and operator in that the fields all have different names (existing engine already handles that)... The only gap I can find that needs addressing is to allow multiple rules on the SAME field, specifically in the suffix/prefix/anything-but-[prefix|suffix].  Do you disagree?

@baldawar @timbray ... Curious of your thoughts... From what I can gather, the OR operator implementation essentially just built "n" separate rules with the same name, with each permutation of "or"s children.  In your view, what is the right way to build the $and operator to allow multiple conditions on the same key... Things I can think of:
1) Modify the parser and the state machines to allow multiple rules on a single input field
2) Modify the parser to to create a separately named secondary rule (think parent-child) with each "and" condition, link it to the parent rule, and do a post-order cleanup after executing to reconcile the parent-child relationship (redact the children returning to the caller, redact the parent if all children aren't present) -- based on his comments on my last PR, I'M guessing @timbray would see this as a hack, though its less impactful/risky to the overall codebase

[...]

PR #212, opened 2025-04-24
Introduce option to detect and fail on path overriding in rule definitions

nicdard:
### Description of changes:

Detect and fail in case of path overriding in rule definition which causes the undefined behavior as described in issue #22. 
Note that in case of OR explicit operators we do not need to fail, as the resulting patterns are being copied and therefore not overridden. 

* Tests are added to `JsonRuleCompiler` as well as`RuleCompiler`.
* The `GenericMachine` class and related builder has been modified to be configured. 
   * By default, the previous behavior is maintained, where the override is happening to avoid breaking compatibility.
* No breaking changes are introduced.

```
<replace this with output from /src/test/software/amazon/event/ruler/Bechmarks.java here.

The benchmark results can be fetched from "Pull request checks -> Java build -> build (ubuntu-X.Y, 8) -> Run benchmarks".>
```

---

### DIFF
--- a/src/main/software/amazon/event/ruler/GenericMachine.java
+++ b/src/main/software/amazon/event/ruler/GenericMachine.java
@@ -798,7 +798,7 @@ public class GenericMachine<T> {
             return this;
         }
 
-        public Builder<M, T> withRuleOverriding(boolean ruleOverriding) {
+        public Builder<M, T> withOverridesForDuplicateRules(boolean ruleOverriding) {
             this.ruleOverriding = ruleOverriding;
             return this;
         }
--- a/src/main/software/amazon/event/ruler/GenericMachine.java
+++ b/src/main/software/amazon/event/ruler/GenericMachine.java
@@ -764,10 +764,42 @@ public class GenericMachine<T> {
          */
         private boolean additionalNameStateReuse = false;
 
+        /**
+         * If true, when encountering the same name path in a rule, the compiler will accept the rule and
+         * override previously encountered patterns with the latest patterns from the compilation of the
+         * last-most occurrence of the name path. For example:
+         * <pre>
+         * {@code
+         *   {
+         *     "a": [1, 2]
+         *     "a": [3, 4]
+         *   }
+         * }
+         * </pre>
+         * Will have the same effect as:
+         * <pre>
+         *   {@code
+         *   {
+         *    "a": [3, 4]
+         *    }
+         * }
+         * </pre>
+         *
+         * When set to false, all rules where paths are overlapping are considered invalid and rejected by the compiler.
+         * For retro-compatibility, rule overriding is by default true.
+         * @see <a href="https://github.com/aws/event-ruler/issues/22">Undocumented duplicate key behaviour for events and rules</a>
+         */
+        private boolean ruleOverriding = true;
+
         Builder() {}
 
         public Builder<M,T> withAdditionalNameStateReuse(boolean additionalNameStateReuse) {
             this.additionalNameStateReuse = additionalNameStateReuse;
+            return this;
+        }
+
+        public Builder<M, T> withRuleOverriding(boolean ruleOverriding) {
+            this.ruleOverriding = ruleOverriding;
             return this;
         }
 
@@ -776,7 +808,7 @@ public class GenericMachine<T> {
         }
 
         protected GenericMachineConfiguration buildConfig() {
-            return new GenericMachineConfiguration(additionalNameStateReuse);
+            return new GenericMachineConfiguration(additionalNameStateReuse, ruleOverriding);
         }
     }
 }
--- a/src/main/software/amazon/event/ruler/GenericMachine.java
+++ b/src/main/software/amazon/event/ruler/GenericMachine.java
@@ -412,7 +412,7 @@ public class GenericMachine<T> {
      */
     public void addRule(final T name, final String json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json).forEach(rule -> addPatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> addPatternRule(name, rule));
         } catch (JsonParseException e) {
             addPatternRule(name, RuleCompiler.compile(json));
         }
@@ -427,7 +427,7 @@ public class GenericMachine<T> {
      */
     public void addRule(final T name, final Reader json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json).forEach(rule -> addPatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> addPatternRule(name, rule));
         } catch (JsonParseException e) {
             addPatternRule(name, RuleCompiler.compile(json));
         }
@@ -442,7 +442,7 @@ public class GenericMachine<T> {
      */
     public void addRule(final T name, final InputStream json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json).forEach(rule -> addPatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> addPatternRule(name, rule));
         } catch (JsonParseException e) {
             addPatternRule(name, RuleCompiler.compile(json));
         }
@@ -457,7 +457,7 @@ public class GenericMachine<T> {
      */
     public void addRule(final T name, final byte[] json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json).forEach(rule -> addPatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> addPatternRule(name, rule));
         } catch (JsonParseException e) {
             addPatternRule(name, RuleCompiler.compile(json));
         }
@@ -472,7 +472,7 @@ public class GenericMachine<T> {
      */
     public void deleteRule(final T name, final String json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json).forEach(rule -> deletePatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> deletePatternRule(name, rule));
         } catch (JsonParseException e) {
             deletePatternRule(name, RuleCompiler.compile(json));
         }
@@ -487,7 +487,7 @@ public class GenericMachine<T> {
      */
     public void deleteRule(final T name, final Reader json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json).forEach(rule -> deletePatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> deletePatternRule(name, rule));
         } catch (JsonParseException e) {
             deletePatternRule(name, RuleCompiler.compile(json));
         }
@@ -502,7 +502,7 @@ public class GenericMachine<T> {
      */
     public void deleteRule(final T name, final InputStream json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json).forEach(rule -> deletePatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> deletePatternRule(name, rule));
         } catch (JsonParseException e) {
             deletePatternRule(name, RuleCompiler.compile(json));
         }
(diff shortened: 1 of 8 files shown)


timbray (commented):
I don't have time to do a thorough review but I don't see anything obviously wrong with the code. The test looks OK, and I hope you've had a careful look at the coverage.  I also didn't see the benchmark output, does this PR have any performance effects?

However, when a rule is supplied via JSON, it makes me very nervous that a JSON object with duplicate keys is accepted. Per RFC8259, such JSON is not interoperable and should not be accepted. Would it be possible to simply reject such JSON rules?

Of course, you could still get conflicting paths by adding rules at the Patterns level, so this is still useful even if you reject JSON with duplicate keys.

nicdard:
@timbray thank you for taking a look! 

> I don't have time to do a thorough review but I don't see anything obviously wrong with the code. The test looks OK, and I hope you've had a careful look at the coverage. 

Test should be covering all the logic, I added enough to test the various cases and corner cases (such the ones described in the issue)

> However, when a rule is supplied via JSON, it makes me very nervous that a JSON object with duplicate keys is accepted. Per RFC8259, such JSON is not interoperable and should not be accepted. Would it be possible to simply reject such JSON rules?

Regarding the above, I think we should still accept it bc of the implicit and operation (I see there is an issue related to the introduction of an new $and operator https://github.com/aws/event-ruler/issues/69), see this example:
https://github.com/aws/event-ruler/issues/22#issuecomment-2827172532

> Of course, you could still get conflicting paths by adding rules at the Patterns level, so this is still useful even if you reject JSON with duplicate keys.

Yes I was thinking that this change just solves all the case where the user could get confused and enter a filter where the semantic is wrong with respect to what was thought to be.


I will run the benchmarks and post the results in this PR

timbray:
> @timbray thank you for taking a look!

First of all, I have to say that I will not approve this PR.  I haven't touched Ruler code in 3 years or more, and I'm not confident enough about not breaking things.   It really needs an insider contributor to look at it.

BTW the former chief maintainer @baldawar has left AWS and moved to Square, so I'm not sure who has ownership of this code on the AWS side.

> Test should be covering all the logic, I added enough to test the various cases and corner cases (such the ones described in the issue)

Not interested in "should be" I would want to hear that (a) you have checked and the coverage % does not go down, and (b) the new code you have written has basically 100% test coverage except for places that obviously don't benefit.

> > However, when a rule is supplied via JSON, it makes me very nervous that a JSON object with duplicate keys is accepted. Per RFC8259, such JSON is not interoperable and should not be accepted. Would it be possible to simply reject such JSON rules?
> 
> Regarding the above, I think we should still accept it bc of the implicit and operation (I see there is an issue related to the introduction of an new $and operator #69), see this example: [#22 (comment)](https://github.com/aws/event-ruler/issues/22#issuecomment-2827172532)

Not sure I understand, but I still don't think Ruler should be accepting broken JSON.

> Regarding the benchmarks, I saw this line in the [PR template](https://github.com/aws/event-ruler/blob/main/.github/PULL_REQUEST_TEMPLATE.md#benchmark--performance-for-source-code-changes):
> 
> > The benchmark results can be fetched from "Pull request checks -> Java build -> build (ubuntu-X.Y, 8) -> Run benchmarks".>
> 
> Could you please approve the workflow to run so that I can get the results. If I run the benchmarks in my machine, I'll get a different result (double or more the speed).

I approved it but it failed because the git actions are out of date, some Ubuntu version thing.  If you want to do another PR to fix that, it is the kind of thing I would approve.

> as expected the performance is quite similar since the change is only affecting the compilation of the filter. Should I post the change where I substitute the direct constructor invocation of `Machine` class with the builder call?

nicdard:
> First of all, I have to say that I will not approve this PR. I haven't touched Ruler code in 3 years or more, and I'm not confident enough about not breaking things. It really needs an insider contributor to look at it.

Thank you for, makes sense no worries.

> BTW the former chief maintainer @baldawar has left AWS and moved to Square, so I'm not sure who has ownership of this code on the AWS side.

I see thanks for the info.

> > Test should be covering all the logic, I added enough to test the various cases and corner cases (such the ones described in the issue)
> 
> Not interested in "should be" I would want to hear that (a) you have checked and the coverage % does not go down, and (b) the new code you have written has basically 100% test coverage except for places that obviously don't benefit.

I measured the coverage, and there is no change in it. all the new branches are covered in the code.

> > > However, when a rule is supplied via JSON, it makes me very nervous that a JSON object with duplicate keys is accepted. Per RFC8259, such JSON is not interoperable and should not be accepted. Would it be possible to simply reject such JSON rules?
> > 
> > 
> > Regarding the above, I think we should still accept it bc of the implicit and operation (I see there is an issue related to the introduction of an new $and operator #69), see this example: [#22 (comment)](https://github.com/aws/event-ruler/issues/22#issuecomment-2827172532)
> 
> Not sure I understand, but I still don't think Ruler should be accepting broken JSON.

I see your point, maybe it can be addressed separately given the above discussion I was mentioning.

> > Regarding the benchmarks, I saw this line in the [PR template](https://github.com/aws/event-ruler/blob/main/.github/PULL_REQUEST_TEMPLATE.md#benchmark--performance-for-source-code-changes):
> > > The benchmark results can be fetched from "Pull request checks -> Java build -> build (ubuntu-X.Y, 8) -> Run benchmarks".>
> > 
> > 
> > Could you please approve the workflow to run so that I can get the results. If I run the benchmarks in my machine, I'll get a different result (double or more the speed).
> 
> I approved it but it failed because the git actions are out of date, some Ubuntu version thing. If you want to do another PR to fix that, it is the kind of thing I would approve.

Thank you! I can see if I have time I can open a new PR :)


> > as expected the performance is quite similar since the change is only affecting the compilation of the filter. Should I post the change where I substitute the direct constructor invocation of `Machine` class with the builder

timbray:
Hey @embano1 is there anyone at AWS who should be looking at this?

embano1:
Pinged some folks internally.

rishibaldawa:
hey @embano1, 👋  I'd recommend starting w @kamouney @scabraha . 

@nicdard will look at this PR later today.

src/main/software/amazon/event/ruler/GenericMachine.java
```diff
@@ -764,19 +764,51 @@ public static class Builder<M extends GenericMachine<T>, T> {
...
+        private boolean ruleOverriding = true;
+
         Builder() {}
 
         public Builder<M,T> withAdditionalNameStateReuse(boolean additionalNameStateReuse) {
             this.additionalNameStateReuse = additionalNameStateReuse;
             return this;
         }
 
+        public Builder<M, T> withRuleOverriding(boolean ruleOverriding) {
```
rishibaldawa, line 801: withOverridesForDuplicateRules might be more clearer.

src/main/software/amazon/event/ruler/JsonRuleCompiler.java
```diff
@@ -343,7 +380,12 @@ private static void writeRules(final List<Map<String, List<Patterns>>> rules,
         if (rules.isEmpty()) {
             rules.add(new HashMap<>());
         }
-        rules.forEach(rule -> rule.put(name, values));
+        rules.forEach(rule -> {
+            if (!withOverriding && rule.containsKey(name)) {
+                throw new IllegalArgumentException("Overriding path '" + name + "' is already defined. Please check that the rule does not list the same path multiple times.");
```
rishibaldawa, line 385: I think we use barf() for error handling. Also, wouldn't recommend putting in the name in the exception message in case there's something sensitive that folks don't want to be logged.

src/main/software/amazon/event/ruler/RuleCompiler.java
```diff
@@ -240,6 +274,9 @@ private static void writeRules(final Map<String, List<Patterns>> rule,
         if (values.isEmpty()) {
             barf(parser, "Empty arrays are not allowed");
         }
+        if (!withOverriding && rule.containsKey(name)) {
+            throw new IllegalArgumentException("Overriding path '" + name + "' is already defined. Please check that the rule does not list the same path multiple times.");
```
rishibaldawa, line 278: Same comments as in the other file.

src/main/software/amazon/event/ruler/JsonRuleCompiler.java
rishibaldawa: Error msg can be `barf("Same path cannot be allowed multiple times")` for consistency w rest of the messages

ghost:
also @nicdard , TY 

Screenshot 2025-05-14 at 5 10 56 PM

src/main/software/amazon/event/ruler/JsonRuleCompiler.java
nicdard: oh thanks good point, I'll use barf as well.
however in terms of sensitive info, I guess barf is worst since it's going to print all the source as a string:

From `JsonProcessingException`:
```java
@Override public String getMessage() {
        String msg = super.getMessage();
        if (msg == null) {
            msg = "N/A";
        }
        JsonLocation loc = getLocation();
        String suffix = getMessageSuffix();
        // mild optimization, if nothing extra is needed:
        if (loc != null || suffix != null) {
            StringBuilder sb = new StringBuilder(100);
            sb.append(msg);
            if (suffix != null) {
                sb.append(suffix);
            }
            if (loc != null) {
                sb.append('\n');
                sb.append(" at ");
                sb.append(loc.toString());
            }
            msg = sb.toString();
        }
        return msg;
    }
```

If it's ok I would state the path in the message, just bc it's more clear which one is the offending path. wdyt?

src/main/software/amazon/event/ruler/GenericMachine.java
nicdard: refactored :)

nicdard:
@rishibaldawa thank you for your review, all the changes are in this commit: https://github.com/aws/event-ruler/pull/212/commits/46aced22a208da2211fbf35d60bf83826c04d526

I will re-order the commits if needed to have the change of version as last. don't know if you do squash and merge than we don't need that. leaving them in this order so you can more easily follow up in the review for now :)

src/main/software/amazon/event/ruler/JsonRuleCompiler.java
ghost: sounds good to me though I'll let AWS make the final call if this is acceptable from a security standpoint.

ghost:
We used to squash and merge. I'd imagine we'd do the same going forward unless there's a good reason.

shikhartanwar:
@rishibaldawa Who would be the right person to reach out to for getting this PR merged?

ghost:
@kamouney should be the one.

shikhartanwar:
> @kamouney should be the one.

@kamouney Could you help us with the approval and merging the PR?

shikhartanwar:
> @kamouney Could you help us with the approval and merging the PR?

Thanks a lot for the approval @kamouney. I see that all checks have passed now. Could you please merge it and generate a release?

shikhartanwar:
> Thanks a lot for the approval @kamouney. I see that all checks have passed now. Could you please merge it and generate a release?

@kamouney Sorry for bothering you multiple times, but can you please generate a latest release/tag that has this fix.

nicdard:
> > Thanks a lot for the approval @kamouney. I see that all checks have passed now. Could you please merge it and generate a release?
> 
> @kamouney Sorry for bothering you multiple times, but can you please generate a latest release/tag that has this fix.

+1  this would allow us to use the new version :)

shikhartanwar:
> > > Thanks a lot for the approval @kamouney. I see that all checks have passed now. Could you please merge it and generate a release?
> > 
> > 
> > @kamouney Sorry for bothering you multiple times, but can you please generate a latest release/tag that has this fix.
> 
> +1 this would allow us to use the new version :)

@kamouney Any update on this please?

kamouney:
> > > > Thanks a lot for the approval @kamouney. I see that all checks have passed now. Could you please merge it and generate a release?
> > > 
> > > 
> > > @kamouney Sorry for bothering you multiple times, but can you please generate a latest release/tag that has this fix.
> > 
> > 
> > +1 this would allow us to use the new version :)
> 
> @kamouney Any update on this please?

Sorry for the delay, I released [V1.8.2 ](https://github.com/aws/event-ruler/releases/tag/v1.8.2)

nicdard:
> > > > > Thanks a lot for the approval @kamouney. I see that all checks have passed now. Could you please merge it and generate a release?
> > > > 
> > > > 
> > > > @kamouney Sorry for bothering you multiple times, but can you please generate a latest release/tag that has this fix.
> > > 
> > > 
> > > +1 this would allow us to use the new version :)
> > 
> > 
> > @kamouney Any update on this please?
> 
> Sorry for the delay, I released [V1.8.2 ](https://github.com/aws/event-ruler/releases/tag/v1.8.2)

Thank you!

PR #244, opened 2026-04-02
Add rulesForJSONEventNonBlocking with linear-time array-consistent matching

fym-rgb:
## Summary

Add a new matching method `rulesForJSONEventNonBlocking()` that provides
array-consistent matching with linear performance regardless of array size.

The existing `rulesForJSONEvent()` can exhibit quadratic performance when
events contain large arrays at paths matching multi-field rules. The new
method avoids this by walking the compiled state machine trie with direct
path lookups (HashMap) instead of scanning all remaining fields per step.

A configuration flag `withStructuredMatching(true)` on Machine.Builder
allows `rulesForJSONEvent()` to use the new implementation.

## Changes

| File | Change |
|------|--------|
| StructuredFinder.java | New: indexed trie walker with early-exit |
| GenericMachine.java | New method + deprecation + config flag |
| GenericMachineConfiguration.java | useStructuredMatching flag |
| ACFinder.java | Field-index optimization in moveFrom() |
| Event.java | Field name range index + getFieldRange() |
| NameState.java | getValueTransitionKeys() |
| SubRuleContext.java | getRuleCount() for early-exit |
| ArrayMembership.java | Package-visible size() and entries() |
| ACMachineTest.java | createMachine() factory for subclass override |
| ACMachineStructuredTest.java | Runs all 71 ACMachineTest cases with structured matching |
| StructuredFinderTest.java | 51 correctness cases + 12 perf scenarios |
| correctness-cases.json | Test data |

## How it works

The existing ACFinder flattens the event into a list of (name, value,
membership) triples, then for each state transition scans all remaining
fields. With N array elements and K rule conditions, this is O(N*K) scans
of O(N) fields = O(N^2*K).

StructuredFinder indexes the event by dotted field path (HashMap), then
walks the trie with direct lookups. Each transition is O(1) instead of
O(N). With early-exit, it stops as soon as all rules are matched.

## Performance

Tested with payloads up to 1.2MB across 28 scenarios including attack
payloads. All times in ms:

| Scenario | N | KB | Before | After |
|----------|---|-----|--------|-------|
| Object array, 2-field rule | 45,000 | 1,121 | OOM | 74 |
| 3-level nested arrays | 380 | 1,125 | OOM | 49 |
| 10-condition rule | 8,000 | 1,177 | OOM | 57 |
| Customer-like 6-field | 130,000 | 1,161 | OOM | 29 |
| Wildcard bomb attack | 45,000 | 1,121 | OOM | 93 |
| 50-rule fan-out attack | 45,000 | 1,121 | OOM | 74 |
| Duplicate values attack | 65,000 | 1,143 | OOM | 68 |

Worst case at 1MB: 118ms (5-condition almost-match, full walk, no early-exit).

## Correctness

- 679 existing tests pass unchanged
- 71 ACMachineTest cases re-run with structured matching (ACMachineStructuredTest)
- 51 new correctness cases comparing old vs new method
- Externally validated against 2,799 test cases across all code paths

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,6 +637,23 @@ NameState. This is accomplished by storing that we already have a NameState for 
 
 Note that it doesn't matter if each addRule uses a different rule name or the same rule name.
 
+#### withStructuredMatching
+Default: false
+
+When set to true, `rulesForJSONEvent()` uses a linear-time matching algorithm instead of the
+default step-queue approach. This is recommended for applications that process events with large
+arrays (thousands of elements) where the default algorithm may exhibit quadratic performance.
+
+The linear-time algorithm (`StructuredFinder`) indexes the event by field path and walks the
+compiled state machine trie with direct HashMap lookups instead of scanning all remaining fields
+per step. It also exits early once all rules in the machine have been matched.
+
+```java
+Machine machine = Machine.builder()
+    .withStructuredMatching(true)
+    .build();
+```
+
 ### addRule()
 
 All forms of this method have the same first argument, a String which provides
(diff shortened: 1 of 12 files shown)


timbray:
Looks clever!

If it has the same semantics, why don't you just replace `rulesForJSONEvent`? Are there any downsides?  If we do need two separate endpoints, the docs should provide clear guidance on when to use one rather than the other.

fym-rgb:
@timbray
Yep - still making some improvements here and there but I guess I still need to talk with others in the team on this before this can be merged. Marking as draft for now.

PR #263, opened 2026-07-30
Fix order-dependent deletion for superset value lists (#260)

### LINKED ISSUE #260: Machine.deleteRule with a superset value list can silently no-op depending on value order
## Describe the bug

Rule value lists have OR semantics (`"x":["a","b"]` matches `a` OR `b`). But `deleteRule` with a value list that is a superset of the rule's can silently fail to delete, depending on which value the delete walk processes first.

Mechanism: in `GenericMachine.deleteStep`, when the candidate sub-rule set trims to empty at a non-terminal key, the walk returns immediately — skipping the remaining value patterns for that key. If the first value processed is one the target rule does not use (so candidates initialize to empty), later values the rule *does* use are never processed, and the rule survives. Because an empty candidate set also means "uninitialized", the early return can't simply be removed — the empty-set overload needs untangling.

Found while reviewing #256; reproduces identically on `main` before that PR (both with exact values and wildcards), so it is a long-standing quirk, not a regression.

## To Reproduce

Deterministic (no iteration-order dependence needed for the exact-value shape):

```java
@Test
public void supersetValueDelete_singleNameState_exacts() throws Exception {
    Machine m = Machine.builder().build();
    m.addRule("r1", "{\"x\":[\"b\"],\"y\":[\"e\"]}");
    m.addRule("r3", "{\"x\":[\"a\"],\"y\":[\"e\"]}");   // makes "a" exist in the machine

    // r1 is reachable via "b", which is in the provided list — but the delete no-ops
    // because "a" is processed first and empties the candidate set.
    m.deleteRule("r1", "{\"x\":[\"a\",\"b\"],\"y\":[\"e\"]}");

    List<String> afterB = m.rulesForJSONEvent("{\"x\":\"b\",\"y\":\"e\"}");
    assertFalse("r1 should be deleted (reachable via \"b\")", afterB.contains("r1"));  // FAILS
    assertTrue("r3 must survive", m.rulesForJSONEvent("{\"x\":\"a\",\"y\":\"e\"}").contains("r3"));
}
```

Note the documented `deleteRule` contract explicitly supports partial deletes (deleting `r1 {a,[1]}` from `r1 {a,[1,2]}`); this is the mirrored case (superset delete list) behaving inconsistently: at a terminal key it works, at a non-terminal key it depends on value order.

## Expected behavior

Delete outcome should be independent of value order within a key: any sub-rule of the named rule reachable via the provided name/values gets its matching registrations removed.

## Notes

- The multi-NameState branch added in #256 already behaves OR-consistently here (it continues to later values); the quirk lives in the single-NameState early-return, which predates #256.
- Fix likely involves replacing the empty-means-uninitialized candidate-set overload with an explicit initialized flag (or sentinel), so an empty trimmed set can skip recursion for that value without aborting the remaining values.

Pybsama:
### Issue #, if available:

Closes #260

### Description of changes:

`Machine.deleteRule` could silently fail when the supplied rule contained a
superset of the stored values. Values within a field have OR semantics, but
the delete path reused a mutable candidate set across those alternatives and
also used an empty set to mean both "not initialized yet" and "no candidates
matched the preceding keys." The result depended on value order.

This change:

- tracks candidate-set initialization explicitly;
- evaluates every value/NameState alternative from the same candidates inherited
  from preceding keys;
- prevents an empty candidate set at a later key from being reinitialized; and
- limits an empty branch to that branch instead of skipping the remaining
  alternatives for the field.

The existing multi-NameState wildcard teardown guard remains unchanged.

Testing:

- Added order-independent reproductions for exact and wildcard value lists.
- Added prior-key and terminal boundary tests to prevent candidate reinitialization.
- Added coverage for deleting independent sub-rules with the same rule name.
- The five focused regressions passed in 20 consecutive runs.
- `mvn -Dtest=MachineTest test`: 91 tests passed.
- `mvn verify`: 772 tests passed; Checkstyle and SpotBugs reported no findings.

Environment:

- Apple Silicon, macOS 26.5.1
- OpenJDK 21.0.12
- Maven 3.9.16

Full comparison command:

```text
./scripts/perf-compare.sh \
  337e8d99026d3f6687f783535f157be8bf4ebbb7 \
  0e0b36545d52b71be24355349a0ce7f314706db0 \
  -- -Dstyle.color=never -Druler.perf.warmup=5 -Druler.perf.measure=10
```

```text
rule_type                     before_eps     after_eps   delta_pct  verdict
---------                     ----------     ---------   ---------  -------
EXACT                             431543        428359       -0.7%  noise (±2.8%)
WILDCARD                          276394        341981      +23.7%  improvement (noise ±3.2%)
PREFIX                            433777        429462       -1.0%  noise (±2.1%)
PREFIX_EIC                        424932        429335       +1.0%  noise (±9.5%)
SUFFIX                            422192        425125       +0.7%  noise (±4.2%)
SUFFIX_EIC                        421414        422822       +0.3%  noise (±2.3%)
EQUALS_IGNORE_CASE                380035        382514       +0.7%  noise (±2.7%)
NUMERIC                           279480        281772       +0.8%  noise (±1.3%)
ANYTHING_BUT                      250883        242854       -3.2%  REGRESSION (noise ±3.0%)
ANYTHING_BUT_IGNORE_CASE          243198        245030       +0.8%  noise (±2.4%)
ANYTHING_BUT_PREFIX               254955        259433       +1.8%  noise (±2.6%)
ANYTHING_BUT_SUFFIX               254234        252456       -0.7%  noise (±2.6%)
ANYTHING_BUT_WILDCARD             231962        282701      +21.9%  improvement (noise ±1.9%)
COMPLEX_ARRAYS                     55038         54984       -0.1%  noise (±1.7%)
```

The only full-run flag was `ANYTHING_BUT`, 0.2 percentage points outside the
script's heuristic noise band. I repeated the whole `ANYTHING_BUT` family with
5 warmup and 15 measurement passes:

```text
rule_type                     before_eps     after_eps   delta_pct  verdict
---------                     ----------     ---------   ---------  -------
ANYTHING_BUT                      254288        235802       -7.3%  noise (±17.4%)
ANYTHING_BUT_IGNORE_CASE          242591        245506       +1.2%  noise (±1.9%)
ANYTHING_BUT_PREFIX               255447        258124       +1.0%  noise (±7.0%)
ANYTHING_BUT_SUFFIX               251671        251312       -0.1%  noise (±4.5%)
ANYTHING_BUT_WILDCARD             232983        234265       +0.6%  noise (±3.1%)
```

No matching-throughput regression was reproducible. This change only affects
rule deletion; the benchmark exercises event matching.

---

### DIFF
--- a/src/main/software/amazon/event/ruler/GenericMachine.java
+++ b/src/main/software/amazon/event/ruler/GenericMachine.java
@@ -251,6 +251,8 @@ public class GenericMachine<T> {
      *  The eventual result will be same as one time deleteRule call with r1 {a, [1,2]}.
      *  So, caller is expected to save its rule expression if want to entirely remove the rule unless deliberately
      *  want to remove partial rule from rule name.
+     *  Values within a key are order-independent; values that reach no rule are skipped while later values are
+     *  still considered.
      *
      * @param name ARN of the rule
      * @param namevals names and values which make up the rule
@@ -309,22 +311,13 @@ public class GenericMachine<T> {
                         : byteMachine.findAllPatterns(pattern);
             }
 
-            if (nextNameStates.size() <= 1) {
-                NameState nextNameState = nextNameStates.isEmpty() ? null : nextNameStates.iterator().next();
-                if (nextNameState != null && deleteStepForNameState(state, key, keyIndex, keys, patterns, ruleName,
-                        deletedKeys, candidateSubRuleIds, pattern, nextNameState, deletedSubRuleIds, nextNameStates)) {
-                    return deletedSubRuleIds;
-                }
-            } else {
-                Set<SubRuleContext> survivingCandidates = new HashSet<>();
-                for (NameState nextNameState : nextNameStates) {
-                    Set<SubRuleContext> incoming = new HashSet<>(candidateSubRuleIds);
-                    deleteStepForNameState(state, key, keyIndex, keys, patterns, ruleName,
-                            deletedKeys, incoming, pattern, nextNameState, deletedSubRuleIds, nextNameStates);
-                    survivingCandidates.addAll(incoming);
-                }
-                candidateSubRuleIds.clear();
-                candidateSubRuleIds.addAll(survivingCandidates);
+            // Values within a key have OR semantics. Each value/NameState branch starts with the same candidates
+            // inherited from the preceding keys.
+            for (NameState nextNameState : nextNameStates) {
+                final Set<SubRuleContext> branchCandidateSubRuleIds = new HashSet<>(candidateSubRuleIds);
+                deleteStepForNameState(state, key, keyIndex, keys, patterns, ruleName,
+                        deletedKeys, branchCandidateSubRuleIds, pattern, nextNameState, deletedSubRuleIds,
+                        nextNameStates);
             }
         }
 
@@ -333,18 +326,18 @@ public class GenericMachine<T> {
 
     // Deletes a single pattern via a single NameState. The teardown guard considers all NameStates the pattern can
     // lead to (nextNameStates), so shared wildcard transitions survive while any other NameState still uses them.
-    private boolean deleteStepForNameState(final NameState state,
-                                           final String key,
-                                           final int keyIndex,
-                                           final List<String> keys,
-                                           final Map<String, List<Patterns>> patterns,
-                                           final T ruleName,
-                                           final List<String> deletedKeys,
-                                           final Set<SubRuleContext> candidateSubRuleIds,
-                                           final Patterns pattern,
-                                           final NameState nextNameState,
-                                           final Set<SubRuleContext> deletedSubRuleIds,
-                                           final Set<NameState> nextNameStates) {
+    private void deleteStepForNameState(final NameState state,
+                                        final String key,
+                                        final int keyIndex,
+                                        final List<String> keys,
+                                        final Map<String, List<Patterns>> patterns,
+                                        final T ruleName,
+                                        final List<String> deletedKeys,
+                                        final Set<SubRuleContext> candidateSubRuleIds,
+                                        final Patterns pattern,
+                                        final NameState nextNameState,
+                                        final Set<SubRuleContext> deletedSubRuleIds,
+                                        final Set<NameState> nextNameStates) {
         // If this was the last step, then reaching the last state means the rule matched, and we should delete
         // the rule from the next NameState.
         final int nextKeyIndex = keyIndex + 1;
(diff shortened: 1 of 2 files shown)


fym-rgb:
Hi @Pybsama 
Thank you for the PR, will review the content in the coming days.

fym-rgb:
Thanks @Pybsama for your contribution. This is a careful fix, and the test design is strong. 

- Your five tests cherry-picked onto unmodified `main` fail 4/5, exactly as they should (the 5th, `testIssue260SupersetDeletePreservesPriorKeyCandidates`, correctly passes on `main` — it pins behavior a naive fix would break).
- `testIssue260TerminalSupersetDoesNotResetPriorKeyCandidates` is a nice catch beyond what #260 documented: on `main`, the empty-set re-init at a *terminal* key **over-deletes** — it removes registrations reachable only via other first-key values. Your fix closes both directions.
- Full `mvn verify` on your branch is green (772 tests), and the five new tests are stable across 20 repeated runs.
- The approach matches the direction we settled in #256: values within a key are OR semantics, so every value/NameState alternative is evaluated from the candidates inherited from the preceding keys. Unifying the single- and multi-NameState paths is a welcome simplification, and I confirmed the wildcard teardown guard is unchanged and nothing reachable from the matching path is touched (so your no-benchmark-regression conclusion holds).

Two changes I'd like before merge:

1. **Simplify the new flag away.** At every call site `candidateSubRuleIdsInitialized` equals `keyIndex != 0` (`deletePatternRule` passes `(0, false)`; the only recursion passes `(keyIndex + 1, true)`). Carrying the same fact twice through the walk invites drift at a future call site. Please drop the parameter from both signatures and test `keyIndex == 0` at the initialization branch — zero behavior change, smaller diff. While you're there: the `new HashSet<>(candidateSubRuleIds)` at the recursion call is now a redundant defensive copy (post-change, `deleteStep` never mutates its parameter — each branch takes its own copy), so it can pass the set directly.

2. **Pin the later-key continuation with a test that actually deletes.** The "empty branch is confined to that branch" behavior is only exercised-with-deletion at the *first* key. Both later-key tests use delete lists that cover no stored sub-rule, so they can't distinguish "continued past the empty branch" from "aborted the key". Please add a variant where the second value at a later key legitimately deletes, e.g.:
   - add `r1` `{"a":["one"],"b":["right"]}` and `other` `{"a":["one"],"b":["missing"]}`
   - delete `r1` with `{"a":["one"],"b":["missing","right"]}` in both value orders
   - assert `r1` is gone, `other` still matches, and the machine is empty after deleting `other`.

Optionally polish these:

- A short comment on why `withAdditionalNameStateReuse(true)` is load-bearing in the two later-key tests (it funnels a key's values into one shared NameState, which is what makes the foreign patterns findable there).
- The new comment's "rather than consuming another branch's result" describes the old behavior; ending at the positive statement reads cleaner.
- A sentence in the `deletePatternRule` javadoc stating the now-guaranteed semantics (values that reach no rule are skipped; the outcome doesn't depend on value order within a key) — #260 existed exactly because this was ambiguous.
- The trailing-blank-line removal at the end of `GenericMachine.java` is unrelated to the fix; restoring it keeps the diff scoped.

When you've made the changes, please mark the PR **ready for review** and we can get this formally reviewed.

---
###### Generative AI Assisted Review

Pybsama:
Addressed the requested changes and marked the PR ready for review:

- Rebased the single fix commit onto current main (a356c6f).
- Removed candidateSubRuleIdsInitialized, derived initialization from keyIndex == 0, and passed the candidate set directly into recursion.
- Added the later-key deletion regression for both [missing,right] and [right,missing], preserving the other rule and confirming final machine teardown.
- Added the NameState-reuse rationale, documented order-independent value semantics, made the branch comment positive, and restored the unrelated trailing blank line.

Fresh verification on JDK 21:
- Issue260 tests: 6 passed
- MachineTest: 98 passed
- mvn verify: 779 passed; Checkstyle and SpotBugs clean
- Benchmarks#CL2*: 2 passed

Current head: d8e4af7ee0b9ef940df7af6526e20efd1982c93d.

fym-rgb (approved):
Thanks @Pybsama. 

Approving. It's been a pleasure reviewing this, thank you for the contribution.

PR #265, opened 2026-08-21
fix: guard forward multi-byte walker against non-InputByte continuation positions

leezachydev:
# Guard the forward multi-byte walker against non-InputByte continuation positions

## Summary
`ByteMachine.extractNextJavaCharacterFromInputCharactersForForwardArrays` reconstructs a multi-byte UTF-8
character by casting the continuation positions (`characters[i + 1]`, `characters[i + 2]`) to `InputByte`
with no type check. When a continuation position holds an `InputMultiByteSet` (which an `equals-ignore-case`
value produces via case folding), the cast throws:

```
java.lang.ClassCastException: class software.amazon.event.ruler.input.InputMultiByteSet
    cannot be cast to class software.amazon.event.ruler.input.InputByte
  at software.amazon.event.ruler.input.InputByte.cast(InputByte.java:19)
  at software.amazon.event.ruler.ByteMachine.extractNextJavaCharacterFromInputCharactersForForwardArrays(ByteMachine.java)
  at software.amazon.event.ruler.ByteMachine.extractNextJavaCharacterFromInputCharacters(ByteMachine.java)
  at software.amazon.event.ruler.ByteMachine.doMultipleTransitionsConvergeForInputByte(ByteMachine.java)
  at software.amazon.event.ruler.ByteMachine.canReuseNextByteState(ByteMachine.java)
  at software.amazon.event.ruler.ByteMachine.addMatchValue(ByteMachine.java)
  at software.amazon.event.ruler.ByteMachine.addMatchPattern(ByteMachine.java)
  at software.amazon.event.ruler.ByteMachine.addPattern(ByteMachine.java)
  at software.amazon.event.ruler.GenericMachine.addStep(GenericMachine.java)
  at software.amazon.event.ruler.GenericMachine.addRule(GenericMachine.java)
```

The failure happens at `addRule` time (rule compilation), so a single valid rule can crash the machine build.

## Reproduction
```java
Machine machine = new Machine();
// A sibling rule seeds the reusable InputMultiByteSet states on the shared "デé" tail...
machine.addRule("sibling",  "{\"k\": [{\"suffix\": {\"equals-ignore-case\": \"b\u30c7\u00e9\"}}]}");
// ...so adding this second rule exercises canReuseNextByteState and throws.
// value = "a" + U+30C7 (デ, 3-byte) + U+00E9 (é, 2-byte)
machine.addRule("offender", "{\"k\": [{\"suffix\": {\"equals-ignore-case\": \"a\u30c7\u00e9\"}}]}");
```
`suffix` + `equals-ignore-case` is a fully supported, documented combination, and non-ASCII rule values are
legitimate, so any caller with such a family of rules hits this deterministically. NOTE: the crash is in the
byte-state-**reuse** path, so it needs a prior rule that seeds a reusable state on the shared multi-byte tail;
a single such rule on a fresh machine does not trip it.

## Verification
Verified against `aws/event-ruler` @ `main` (Apache Maven 3.9.9, Corretto 17): the added test fails on
unpatched `main` with the exact stack below and passes after the fix; the full existing suite stays green
(768 tests, 0 failures, 0 errors).
```
java.lang.ClassCastException: InputMultiByteSet cannot be cast to InputByte
  at InputByte.cast(InputByte.java:19)
  at ByteMachine.extractNextJavaCharacterFromInputCharactersForForwardArrays(ByteMachine.java:939)
  at ByteMachine.doMultipleTransitionsConvergeForInputByte(ByteMachine.java:857)
  at ByteMachine.canReuseNextByteState(ByteMachine.java:822)
  at ByteMachine.addMatchValue(ByteMachine.java:765)
```

## Relationship to #248
#248 ("Guard backwards-array walker against non-InputByte sentinels") added `isByte` guards to
`isContinuationByte` and to `extractNextJavaCharacterFromInputCharactersForBackwardsArrays` (it stops at the
first non-byte and reconstructs from the bytes gathered). It did **not** guard the *forward* walker, which
still performs the unchecked casts. This PR applies the same guard to the forward walker, mirroring #248's
behavior exactly.

## Fix
In `extractNextJavaCharacterFromInputCharactersForForwardArrays`, check each continuation position with a new
bounds-and-type-checked helper `isByteAtIndex(...)` before casting; if a continuation position is not an
`InputByte`, stop and reconstruct from the bytes gathered so far (identical philosophy to the backward walker).

This method is only reached from `doMultipleTransitionsConvergeForInputByte`, which gates the byte-state-reuse
**optimization**. A truncated reconstruction can only cause the optimization to be skipped (a slightly larger
machine); it cannot change match results. The added regression test asserts matching is preserved.

## Testing
- New regression test builds the offending rule and asserts (a) no `ClassCastException`, (b) case-insensitive
  suffix matching still works (exact-case and ASCII case-folded values match; a non-suffix value does not).
- Existing `ByteMachineTest` / `MachineTest` suites for suffix, equals-ignore-case, wildcard, and multi-byte
  patterns must remain green (behavior-preserving change).

## Notes for downstream consumers
This is a net-new
fix: it is NOT resolved by any released version through v2.0.1 (v2.0.0's #248 only covered the backward
walker). Consumers must take the release that contains this change; a version bump within the 2.0.x line prior
to it will not help.

---

### DIFF
--- a/src/main/software/amazon/event/ruler/ByteMachine.java
+++ b/src/main/software/amazon/event/ruler/ByteMachine.java
@@ -933,12 +933,31 @@ class ByteMachine {
         if (firstByte >= MIN_FIRST_BYTE_FOR_ONE_BYTE_CHAR && firstByte <= MAX_FIRST_BYTE_FOR_ONE_BYTE_CHAR) {
             return new String(new byte[] { firstByte } , StandardCharsets.UTF_8);
         } else if (firstByte >= MIN_FIRST_BYTE_FOR_TWO_BYTE_CHAR && firstByte <= MAX_FIRST_BYTE_FOR_TWO_BYTE_CHAR) {
+            // A continuation position can hold a non-byte InputCharacter (e.g. an
+            // InputMultiByteSet from an equals-ignore-case value). Stop before
+            // casting to InputByte, as the backward walker does.
+            if (!isByteAtIndex(characters, i + 1)) {
+                return new String(new byte[] { firstByte }, StandardCharsets.UTF_8);
+            }
             return new String(new byte[] { firstByte, InputByte.cast(characters[i + 1]).getByte() },
                     StandardCharsets.UTF_8);
         } else {
+            if (!isByteAtIndex(characters, i + 1)) {
+                return new String(new byte[] { firstByte }, StandardCharsets.UTF_8);
+            }
+            if (!isByteAtIndex(characters, i + 2)) {
+                return new String(new byte[] { firstByte, InputByte.cast(characters[i + 1]).getByte() },
+                        StandardCharsets.UTF_8);
+            }
             return new String(new byte[] { firstByte, InputByte.cast(characters[i + 1]).getByte(),
                     InputByte.cast(characters[i + 2]).getByte() }, StandardCharsets.UTF_8);
         }
+    }
+
+    // Bounds-checked byte test so a non-byte continuation position never reaches
+    // InputByte.cast() in the forward multi-byte walker.
+    private static boolean isByteAtIndex(InputCharacter[] characters, int i) {
+        return i >= 0 && i < characters.length && isByte(characters[i]);
     }
 
     // Like findPattern, but returns all NameStates a pattern leads to. A shared wildcard can produce one NameState per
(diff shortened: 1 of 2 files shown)


fym-rgb (approved):
Lgtm, added some test coverage.

PR #55, opened 2022-11-10
Speeding up Array Consistency Check by using sorted int-int AVL Tree 

baldawar:
### Description of changes:

This a "I'm throwing this out for everyone else and see if it sticks" kind of PR. 

While digging around for bottlenecks for https://github.com/aws/event-ruler/pull/54, I realized that we had a home-grown Int-Int map and its showing up as a bottleneck (for past context, we used to have a HashMap<Integer, Integer> there and moving to Int-Int map helped improved performance from 1735.6 events/sec to 2645.0 events/sec. The code has since improved but isn't at the same level as rest of the matchers. 

Moving to using a AVL Tree for checking array membership leads to a significant boost from `2821.3 events/sec` to `5275.7 events.sec`. We have the option to gain some more performance if we stop the cloning behaviour in `ArrayMembership(...)` but I've left that as a future exercise (will create a issue for it if this gets merged).

The biggest downside is that its introducing a brand new dependency with lots of classes / structures that we aren't really using. There may be some use for bi-directional iterators or Int2Object Maps in future, but I'm unclear on the benefits for them. so for now, I'd like everyone to assume that we take this new dependency on fastutils-core, it'll be purely for the performance gains for array consistency. 

We can also explore modifying our  own IntIntMap, but looking a how fastutils-core did it, lots of complex edge-cases to solve. 

Note: Apologies for any typos, I sent this PR out between few meetings

#### Benchmark / Performance (for source code changes):

Before 

```
Reading citylots2
Read 213068 events
EXACT events/sec: 193346.6
WILDCARD events/sec: 123374.6
PREFIX events/sec: 189225.6
SUFFIX events/sec: 213068.0
EQUALS_IGNORE_CASE events/sec: 160322.0
NUMERIC events/sec: 111670.9
ANYTHING-BUT events/sec: 123517.7
**COMPLEX_ARRAYS events/sec: 2821.3
PARTIAL_COMBO events/sec: 74028.2
COMBO events/sec: 2517.7**
Reading citylots2
Read 213068 events
Finding Rules...
Lots: 10000
Lots: 20000
Lots: 30000
Lots: 40000
Lots: 50000
Lots: 60000
Lots: 70000
Lots: 80000
Lots: 90000
Lots: 100000
Lots: 110000
Lots: 120000
Lots: 130000
Lots: 140000
Lots: 150000
Lots: 160000
Lots: 170000
Lots: 180000
Lots: 190000
Lots: 200000
Lots: 210000
[...]

PR #271, opened 2026-09-03
docs: add a policy for taking on new dependencies

### LINKED ISSUE #57: Add standard for taking a new dependency
## What is your idea?

Follow up from https://github.com/aws/event-ruler/pull/55#discussion_r1021015618

> > Do we have a standard about adopting dependencies? I remember doing quite a bit of work inside AWS to convince myself/others that a dependency was acceptable. 
> ...
> We don't have a standard  today

We can use this issue to aggregate opinions and prior art from similar libraries. 

## Would you be willing to make the change? 

No

## Additional context

We can update https://github.com/aws/event-ruler/blob/main/CONTRIBUTING.md with some standard guidance on when it is okay to take a new dependency.

wylovelyi:
### Issue #, if available:

Fixes #57

### Description of changes:

Adds `DEPENDENCIES.md`, a written standard for taking on new dependencies. The project had none — #57 was filed by a maintainer straight out of the review discussion on #55.

The checklist is drawn from the concerns actually raised in that discussion rather than invented from scratch:

- **Prefer writing the small amount of code yourself.** A library that drags in a large class hierarchy to replace one small data structure is a bad trade however good the library is.
- **Establish that the payoff is common enough to matter.** A dependency is easier to justify when it improves a path users hit routinely.
- **Show the benefit with numbers, and say where they came from** — `scripts/perf-compare.sh`, naming the benchmark. A number without a named benchmark cannot be checked (this exact question came up in #55).
- **Report the cost, not just the benefit** — classes, size, transitive dependencies. "It is a popular library" is not a measurement.
- **Account for the ongoing governance and security burden**: CVE tracking, patching, upgrades across the transitive tree. A dependency is a commitment, not a one-off import.
- **Record the decision in the PR** so the rationale is findable later and the next person does not re-litigate it.

The doc also states where the project stands today (three compile-scope dependencies: jackson-databind, fastdoubleparser, jsr305), so reviewers have the current baseline in view.

`CONTRIBUTING.md` gains one line in the pre-PR checklist pointing at it. That is the only change to an existing file.

N/A — documentation-only change, no source code modified.

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,7 @@ Contributions via pull requests are much appreciated. Before sending us a pull r
 1. You are working against the latest source on the *main* branch.
 2. You check existing open, and recently merged, pull requests to make sure someone else hasn't addressed the problem already.
 3. You open an issue to discuss any significant work - we would hate for your time to be wasted.
+4. If your change adds or upgrades a dependency, you have walked through the checklist in [DEPENDENCIES.md](DEPENDENCIES.md).
 
 To send us a pull request, please:
 
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -1,0 +1,69 @@
+# Dependency Policy
+
+Ruler is a low-level Java library published to Maven Central. Every dependency
+we take on is inherited by every application that depends on us, so the cost of
+adding one is paid mostly by other people. Please read this before opening a
+pull request that adds or materially upgrades a dependency.
+
+## Where we are today
+
+The library currently declares three compile-scope dependencies:
+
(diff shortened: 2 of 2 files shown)


fym-rgb:
Thank you for the careful synthesis. I'm going to close this one, and the reason is not about the text. 

A dependency standard is the maintainers' own statement of what we will accept and what we commit to carrying afterwards, affecting every artifact we ship, on behalf of every consumer downstream of us. Under [GOVERNANCE.md](https://github.com/aws/event-ruler/blob/main/GOVERNANCE.md) that kind of policy is the maintainers' to set. 

So we'll write it ourselves; it will also carry decisions only we can make, like where the bar sits, what we already ship and why, and how upgrades and vulnerability reports are handled. 

Thanks again for contributing.

PR #272, opened 2026-09-03
Add benchmark for MachineComplexityEvaluator (issue #26)

### LINKED ISSUE #26: Benchmark test for MachineComplexityEvaluator
Add a benchmark test for MachineComplexityEvaluator to detect performance regressions in machine complexity evaluation.

@[baldawar]
don't think its worth blocking on, but probably worth adding a benchmark test around this class to ensure evaluator doesn't grossly become inefficient after refactoring. for now, could be github issue that we address later.

@[jonessha]
This is a good idea. Note that we partially have this through testEvaluateBeyondMaxComplexity in MachineComplexityEvaluatorTest. This makes sure that we cap evaluation at max complexity. If we don't, the test will more or less spin forever. So this provides some protection. Still a good idea to have a dedicated benchmark test though. Some consumers of this library might run with a really high max complexity and be really sensitive to latency regressions.

Discussion from https://github.com/aws/event-ruler/pull/18

wylovelyi:
### Issue #, if available:

Closes #26

### Description of changes:

Adds `MachineComplexityEvaluatorBenchmarks`, a focused benchmark for `MachineComplexityEvaluator`, to help detect performance regressions in machine complexity evaluation.

Why a standalone class (rather than folding into `StableBenchmarks`): `StableBenchmarks` measures `rulesForJSONEvent` throughput (events/sec) over the `citylots2` dataset — a different axis. Complexity evaluation is the latency of a *single* `evaluateComplexity` call on one machine, so it does not fit the "build machine + scan dataset" model. This class is the complexity-eval counterpart, in the same spirit as the `@Ignore`'d manual timing tests in `Benchmarks.java`.

What it provides:

- Two `@Ignore`'d manual timing benchmarks (run by un-ignoring, consistent with `Benchmarks.java`):
  - `benchmarkQuaminaExploder` — the historical "Quamina exploder" wildcard pattern set (33 patterns, complexity 45).
  - `benchmarkHighMaxComplexity` — a high `maxComplexity` (10,000,000) single-pattern path, the latency-sensitive configuration called out in the issue.
- One non-ignored CI guard `complexityEvaluationStaysBounded` that fails the build if evaluation becomes grossly inefficient (e.g. the `maxComplexity` cap is defeated), so a regression does not go unnoticed.

Run manually with:

```
mvn test -Dtest=MachineComplexityEvaluatorBenchmarks
```

(Un-ignore the two `@Ignore`'d methods to capture their timings; the CI guard runs by default.)

This is a test-only change, so there is no source-code performance impact. The added benchmarks establish a baseline for `MachineComplexityEvaluator.evaluateComplexity` on a dev box:

```
MachineComplexityEvaluator benchmark (Quamina exploder, 33 patterns, maxComplexity=100):
  10656 us/eval
MachineComplexityEvaluator benchmark (long single wildcard pattern, maxComplexity=10_000_000):
  9179 us/eval
```

These numbers are machine-specific and are meant to be compared across revisions to spot regressions, not as absolute targets. Local verification: the module compiles, the non-ignored guard passes, and `checkstyle`/`spotbugs` report no issues.

---

### DIFF
--- a/src/test/software/amazon/event/ruler/MachineComplexityEvaluatorBenchmarks.java
+++ b/src/test/software/amazon/event/ruler/MachineComplexityEvaluatorBenchmarks.java
@@ -1,0 +1,143 @@
+package software.amazon.event.ruler;
+
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Benchmarks for {@link MachineComplexityEvaluator}.
+ *
+ * <p>Most methods here are marked {@code @Ignore} because their wall-clock numbers depend on the
+ * hardware they run on, so they are not part of the normal unit-test suite. After any non-trivial
+ * change to the complexity-evaluation code, re-run them manually (e.g. with
+ * {@code mvn test -Dtest=MachineComplexityEvaluatorBenchmarks}) and compare the printed timings
+ * against the previous run to detect performance regressions, in the same spirit as
+ * {@link Benchmarks}.</p>
+ *
+ * <p>The one non-ignored test asserts a generous upper bound on a realistic worst-case machine. It
+ * is the CI-runnable half of "detect performance regressions in machine complexity evaluation": a
+ * refactor that made evaluation grossly inefficient (for example, by defeating the
+ * {@code maxComplexity} cap) would blow past the bound and fail the build instead of going
+ * unnoticed. It targets the latency path that latency-sensitive consumers care about, as discussed
+ * in <a href="https://github.com/aws/event-ruler/issues/26">issue #26</a>.</p>
+ */
+public class MachineComplexityEvaluatorBenchmarks {
+
+    // Generous bound: a correct, complexity-capped evaluator finishes a realistic worst-case machine
+    // in well under this even on slow CI hardware. A regression that makes evaluation blow up will
+    // exceed it.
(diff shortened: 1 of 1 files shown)


fym-rgb:
Thank you for picking up #26 — a benchmark for `MachineComplexityEvaluator` has been on the list since 2022, and the
Quamina exploder set is the right first machine to time.

I'm going to close this in favor of #273, which I've just opened from the maintainer side, and I want to explain why
rather than ask you to rework yours: the two evaluator regressions this project actually shipped fixes for last week
(#270: recursing once per exact-match list value, and never following `{"exists": false}` edges) both live at the
`NameState` level, so a benchmark that times a bare `ByteMachine` cannot see either of them. #273 builds every
machine through `Machine.addRule`, covers those two shapes plus the leading-`*` pattern sets whose uncapped
evaluation is exponential in the pattern count (the "really high max complexity" case the issue mentions), pins the
reported complexity and ByteMachine walk counts deterministically in every build, and keeps the timings in a class
gated like `StableBenchmarks` rather than behind `@Ignore`. The 5-second CI assertion here times a machine whose
evaluation takes tens of milliseconds and whose complexity (45) sits below the cap it is given (100), so it cannot
notice a defeated cap; #273 bounds capped evaluation on machines whose uncapped walk takes seconds or minutes, which
is the property a consumer at a production cap depends on.

Your dependency-policy PR (#271) is separate and I'll review it on its own. Thanks again for contributing.

Issue #269, opened 2026-09-01
MachineComplexityEvaluator re-walks downstream machines once per exact-match list value: evaluation cost is the product of list sizes


fym-rgb:
## Describe the bug

`MachineComplexityEvaluator.evaluate(ByteState)` recurses into next NameStates once per
`ByteMatch` — that is, once per value of an exact-match list — with no memoization and no
dedupe of recursion targets
([MachineComplexityEvaluator.java#L104-L114](https://github.com/aws/event-ruler/blob/bfb9a5659cb0b8a792ed4ede556c206d63dc3f83/src/main/software/amazon/event/ruler/MachineComplexityEvaluator.java#L104-L114)).
All values of one key's list lead to the same next NameState, so for a rule whose sorted key
chain carries exact-match lists of sizes L1, L2, ..., Lk ahead of a wildcard-bearing key, the
wildcard key's ByteMachine is re-walked **L1 × L2 × ... × Lk** times. Each walk of a machine
with leading-`*` patterns allocates a fresh `getMatchesAccessibleFromEachTransition` map over
the whole NFA (~12-14 MB for four realistic leading-`*` patterns), so evaluation cost explodes
multiplicatively while the reported complexity value — and the actual runtime matching cost the
metric models — stays small.

Rule construction is not the problem: `addRule` for the reproduction below costs ~73 ms /
~3.6 MB (value lists collapse to one NameState per key since #88). The blowup is entirely
inside `evaluateComplexity`, which callers typically run at rule-creation time to gate
overly-complex wildcard rules — the gate itself is the expensive part.

The recursion into next NameStates exists since #62 (v1.1.0).

## To Reproduce

Rule shape: CloudTrail-style pattern with exact-match lists of sizes 2 × 50 × 21 ahead of an
`anything-but` `wildcard` set of four leading-`*` patterns on a later-sorted key (synthetic
values; any values of these cardinalities behave identically):

```json
{
  "detail": {
    "errorCode": ["AccessDenied", "UnauthorizedOperation"],
    "eventName": ["CreateResource00", "...", "DeleteResource24"],
    "eventSource": ["service00.amazonaws.com", "...", "service20.amazonaws.com"],
    "userIdentity": {
      "arn": [{
        "anything-but": {
          "wildcard": ["*kiwi*", "*BASE_QuxB*", "*Delta*", "*a/example-lake-abc98765432*"]
        }
      }]
    }
  },
  "detail-type": ["AWS API Call via CloudTrail"]
}
[...]

PR #273, opened 2026-09-04
test: Add a complexity-evaluator benchmark and regression guard (issue #26)

### LINKED ISSUE #26: Benchmark test for MachineComplexityEvaluator

fym-rgb:
### Issue #, if available:

Fixes #26

### Description of changes:

`MachineComplexityEvaluator` had no benchmark and no cost guard: the two defects fixed in #270 — recursing once per
exact-match list value instead of once per next NameState, and never following `{"exists": false}` edges — each got a
value-level test, and the suite's two timed evaluations cover a capped bare `ByteMachine` and a numeric-only rule; nothing
bounded a wildcard machine's walk across next NameStates and absent-key edges or pinned that the cap holds there, which is
how both stayed invisible until a production rule was measured. This adds a shared corpus of machines whose evaluation cost is known, a
regression test that runs in every build, and a gated benchmark. Test scope only; no `src/main` change.

**`MachineComplexityEvaluatorCorpus`** — the machines, each built through `Machine.addRule` so an evaluation walks the
NameState graph (value transitions, exact-match lists ahead of a wildcard key, absent-key edges) and not only one
`ByteMachine`; a benchmark over a bare `ByteMachine` cannot see either #270 defect. Each entry carries the complexity
and the ByteMachine walk count an evaluation reports, and a tier saying how expensive the uncapped evaluation is:

| machine | complexity | uncapped evaluation |
|---|---|---|
| Quamina exploder (32 wildcards; the set already in `MachineComplexityEvaluatorTest`) | 45 | 32 ms |
| four leading-`*` exclusions as one anything-but set / as four plain wildcards | 7 / 12 | 13 ms / 13 ms |
| exact-match lists of 2 x 50 x 21 values ahead of the anything-but set (the production shape behind #269) | 7 | 14 ms, 4 walks |
| one wildcard `*a*a*…*a*`, `*a` repeated 50 / 100 times | 101 / 201 | 5 ms / 20 ms |
| the 100-repeat machine behind an `{"exists": false}` key | 201 | 19 ms |
| leading-`*` pattern sets `*<tok>*` of 4 / 8 / 12 / 16 patterns | 12 / 24 / 36 / 48 | 2 ms / 0.13 s / 5.3 s / ~3 min and ~7 GB live |
| a 4,045-character wildcard (2,022 repeats) behind an absent key — a rule at the scale of the 4,096-character pattern limit | 4,045 | 15 s |

(Timings from the new benchmark on a 16-core Linux host, Corretto 17.) The leading-`*` sets are the case issue #26's
"consumers with a really high max complexity" would hit: with the cap far above the true value, the walk visits every
reachable subset of the trailing-`*` states and grows about 2.5x per added pattern; capped at 11, the 12- and 16-pattern
sets return at the walk's first step (0.2-0.5 ms) because the shared leading star alone reaches every pattern.

The corpus's `WalkCountingEvaluator` replaces the private copy `MachineComplexityEvaluatorWalkCountTest` held (that test
now imports it; nothing else in it changes).

**`MachineComplexityEvaluatorRegressionTest`** runs in every build, one JUnit test per (machine, cap) — every machine
under a cap of 11 and every fast machine uncapped, 21 tests in about a second, each named like
`[leading-star-set-12, cap 11]`. Each asserts the reported complexity (the lesser of the cap and the true value), the
ByteMachine walk count, and a 5 s bound on the evaluation, under a 120 s JUnit timeout. The machines whose uncapped walk
takes seconds or minutes are evaluated capped only — the walk stops the moment a position set reaches the cap — so a
change that defeats the cap fails the build by the bound or the timeout. Measured after warm-up: 0.08-14 ms capped,
2-32 ms uncapped; the slowest fast machine takes about 300 ms in a cold JVM, and the uncapped bound is taken on a second
evaluation after the first has paid the JIT cost.

Red-proofs, each by editing main code, running the test and reverting: skipping the absent-key edge walk fails the 3
tests over the hidden machines (they report 0); restoring the pre-#270 once-per-ByteMatch recursion walks the production
shape 2,203 times instead of 4 (both its tests fail on the walk pin); disabling the `size >= maxComplexity` return fails
the capped bound of the 12-pattern set at 6.0 s and of the 4,045-character machine at 17.2 s.

**`MachineComplexityEvaluatorBenchmarks`** reports build time, evaluation time and allocation per machine, capped and
uncapped, and asserts the same pins on every row. Gated behind `-Druler.perf.run=true` exactly like `StableBenchmarks`,
taking the same `warmup` / `measure` / `only` / `csv` properties plus `-Druler.perf.heavy` (`false` skips the
seconds-long uncapped evaluations, `true` — the default — runs them once, `all` also runs the minutes-long 16-pattern
set, which needs about 7 GB of live heap). The default run takes about half a minute and fits a 2 GB heap. README's
benchmark section documents both classes.

Verification: `mvn verify` green on Java 17 (818 tests, 0 failures; checkstyle and spotbugs clean); the four touched test
files compile with `javac --release 8`. No JMH class is added, so CI's benchmark step is unchanged.

```
Test-only change; no source-code performance impact. Default benchmark run at this revision (3 warmup, 10 measured):

machine                              cap       complexity  walks  passes   build_us   median_us   alloc_kb
quamina-exploder                     11                11      1      10     1415.8       570.6      143.3
quamina-exploder                     uncapped          45      1      10     1415.8     31707.4    18806.2
leading-star-anything-but-set        11                 7      1      10      202.9     20112.7    13768.1
leading-star-anything-but-set        uncapped           7      1      10      202.9     12905.7    13768.1
leading-star-plain-wildcards         11                11      1      10      138.4      9904.5    10351.9
leading-star-plain-wildcards         uncapped          12      1      10      138.4     13349.2    13971.4
value-lists-then-anything-but        11                 7      4      10      986.7     15005.4    13907.8
value-lists-then-anything-but        uncapped           7      4      10      986.7     13479.5    13907.3
multi-star-50                        11                11      1      10      207.4       319.4      210.0
multi-star-50                        uncapped         101      1      10      207.4      5264.2     4307.3
multi-star-100                       11                11      1      10      311.6       619.5      357.5
multi-star-100                       uncapped         201      1      10      311.6     19955.3    15643.8
multi-star-100-behind-absent-key     11                11      1      10      289.2       595.8      367.1
multi-star-100-behind-absent-key     uncapped         201      1      10      289.2     19009.7    15636.5
leading-star-set-4                   11                11      1      10       99.6      1857.1     2090.9
leading-star-set-4                   uncapped          12      1      10       99.6      2095.1     2269.6
leading-star-set-8                   11                11      1      10      113.3        76.3       71.3
leading-star-set-8                   uncapped          24      1      10      113.3    127426.5   112466.9
leading-star-set-12                  11                11      1      10      172.5       181.0      143.7
leading-star-set-12                  uncapped          36      1       1      172.5   5327849.0  3913383.7
leading-star-set-16                  11                11      1      10      261.6       454.3      247.0
multi-star-2022-behind-absent-key    11                11      1      10     7267.5     13228.7     6000.0
multi-star-2022-behind-absent-key    uncapped        4045      1       1     7267.5  15393435.9  5829498.7
```

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -940,7 +940,7 @@ Events are processed at over 220K/second except for:
 
 ### Running the benchmarks
 
-Two benchmark harnesses are available; both operate on the same `citylots2` dataset and the same fourteen rule types.
+Two matching-throughput harnesses are available; both operate on the same `citylots2` dataset and the same fourteen rule types. The complexity evaluator has its own pair of classes, described at the end of this section.
 
 **[`Benchmarks.java`](src/test/software/amazon/event/ruler/Benchmarks.java)** — single-shot per rule type. Fast, but JVM warmup cost makes single-pass variance between runs of the same code routinely hit 10-20%. Good for eyeballing a large change; not reliable for detecting small regressions.
 
@@ -1008,6 +1008,24 @@ mvn test -Dtest=StableBenchmarks -Druler.perf.run=true \
 ```
 
 `StableBenchmarks` is gated off by default (`-Druler.perf.run=true` required) so it doesn't run in CI. For publication-grade numbers, use the [JMH benchmarks in `src/test/.../jmh/`](src/test/software/amazon/event/ruler/jmh) — those measure steady-state throughput under a full JMH harness.
+
+#### Complexity evaluation
+
+`MachineComplexityEvaluator` has its own pair of test classes, over a shared corpus of machines whose evaluation cost is known ([`MachineComplexityEvaluatorCorpus.java`](src/test/software/amazon/event/ruler/MachineComplexityEvaluatorCorpus.java): the Quamina exploder set, leading-star anything-but sets, exact-match value lists ahead of a wildcard key, one multi-star wildcard up to the 4,096-character pattern limit, two machines hidden behind an `{"exists": false}` key, and leading-star pattern sets whose uncapped evaluation is exponential in the pattern count). Every machine is built through `Machine.addRule`, so an evaluation walks the NameState graph and not only one `ByteMachine`.
+
+**[`MachineComplexityEvaluatorRegressionTest.java`](src/test/software/amazon/event/ruler/MachineComplexityEvaluatorRegressionTest.java)** runs in every build, once per machine and cap: it pins the value every machine reports under a cap of 11 and the uncapped complexity of every fast machine, pins the number of `ByteMachine` walks each evaluation makes, and bounds each evaluation's time — the machines whose uncapped walk takes seconds or minutes are evaluated capped only, so a change that defeats the cap fails the build.
+
+**[`MachineComplexityEvaluatorBenchmarks.java`](src/test/software/amazon/event/ruler/MachineComplexityEvaluatorBenchmarks.java)** reports build time, evaluation time and allocation per machine, capped and uncapped, and is gated like `StableBenchmarks`:
+
+```
+# Default: 3 warmup + 10 measured evaluations per row, except the seconds-long machines' uncapped row (once);
+# the minutes-long machine runs capped only. About half a minute, and a 2 GB heap suffices
+mvn test -Dtest=MachineComplexityEvaluatorBenchmarks -Druler.perf.run=true
+
+# Same knobs as StableBenchmarks (-Druler.perf.warmup, -Druler.perf.measure, -Druler.perf.only, -Druler.perf.csv), plus:
+#   -Druler.perf.heavy=false   skip the seconds-long uncapped evaluations
+#   -Druler.perf.heavy=all     also run the minutes-long one (about 7 GB of live heap: -DargLine=-Xmx12g)
+```
 
 ### Suggestions for better performance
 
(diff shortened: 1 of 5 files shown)


timbray:
Haha "Quamina Exploder".

Try running it with a couple thousand patterns…

(Going to have to return the favor by stealing a Ruler stress-test for Q)

fym-rgb:
@timbray Claude always has its own very interesting way of naming things :)
Hope you wont mind this blatant "ip infringement"

PR #277, opened 2026-09-08
fix: Reject content after the rule's root object

### LINKED ISSUE #276: Rule compilers accept trailing content after the rule object
`JsonRuleCompiler`, `RuleCompiler` and `RuleCompiler.ListBasedRuleCompiler.flattenRule` stop
reading the input once the rule's root object closes. Anything after that closing brace is never
examined, so `check()` reports these rules as valid, `compile()` returns the rule that precedes
the junk, and `Machine.addRule` / `Ruler.matchesRule` accept and match them:

```
{"orderId":[{"exists":true}]}}            <- extra }
{"orderId":[{"exists":true}]}]            <- extra ]
{"orderId":[{"exists":true}]},
{"orderId":[{"exists":true}]} garbage
{"orderId":[{"exists":true}]}{"x":[1]}    <- a second object, silently dropped
```

Reproduced on `main` at b877fa9 (2.1.0). For each input above, `JsonRuleCompiler.check(rule)`
and `RuleCompiler.check(rule)` return `null`, `new Machine().addRule("r", rule)` succeeds and
`rulesForJSONEvent("{\"orderId\":\"abc\"}")` returns `[r]`, and `Ruler.matchesRule(event, rule)`
returns `true`. Jackson itself rejects every one of them on the next `nextToken()` call
(`Unexpected close marker '}': no open Object to close`, and so on) — the compilers just never
make that call.

**Cause.** `doCompile` (both compilers) and `doFlattenRule` do: `nextToken()` must be
`START_OBJECT` → `parseObject(...)` up to the matching `END_OBJECT` → `parser.close()`. Jackson's
streaming parser tokenizes only what it is asked for, so the bytes after the root object are never
read. Present since the first commit.

**Expected.** `check()` returns an error and `compile()` / `addRule` throw for any non-whitespace
content after the rule object. Trailing whitespace stays valid.

**Fix.** After the root object is consumed, one more `nextToken()`: it throws Jackson's own error
for a malformed trailer, and a non-null token (a second value) gets a compiler error such as
"Filter must be a single JSON object; found content after it". One helper called from the three
sites.

Note for consumers: this narrows the accepted input. Rules that were stored with a trailing brace
compile today and would fail to compile after the fix, so anyone rebuilding machines from stored
rules should scan them with the new `check()` before upgrading.

fym-rgb:
Fixes #276.

`JsonRuleCompiler.doCompile`, `RuleCompiler.doCompile` and `ListBasedRuleCompiler.doFlattenRule` all read the rule's root object and then close the parser. Jackson's streaming parser tokenizes only what it is asked for, so the bytes after the root object were never examined: a trailing `}`, `]`, `,`, a bare word, or a whole second `{...}` compiled as if the rule were valid, and `Machine.addRule` / `Ruler.matchesRule` accepted and matched such rules.

**Commit 1 — the fix.** One package-private helper, `JsonRuleCompiler.requireEndOfInput(parser)`, called from the three sites after the root object is consumed: the next token must be end-of-input. A malformed trailer makes `nextToken()` throw Jackson's own `JsonParseException` (`Unexpected close marker '}': no open Object to close`, and so on); a second well-formed value returns a token and is refused with `Filter must be a single JSON object; found content after it`. Trailing whitespace stays valid — it is part of a JSON text per RFC 8259 and carries nothing.

**Commit 2 — a migration option for machines built from a legacy store.** `Machine.builder().withTrailingContentIgnored(true)` makes `addRule` and `deleteRule` read a rule the way releases before 2.2.0 did (stop at the root object's closing brace). It exists because a consumer that rebuilds a machine from stored rules may hold rules that were accepted under the old reading, and a rejected stored rule fails the whole machine build; the option lets such a consumer keep serving while it re-validates the store. The default is strict, and `check()` never takes the option, so `check()` is the tool for finding those rules. README documents the option beside `withStructuredMatching`.

**Tests.** Commit 1: five new tests across `JsonRuleCompilerTest`, `RuleCompilerTest` and `RulerTest` — every `check()` overload, `compile()`, `flattenRule`, `Machine.addRule`, `Ruler.matchesRule` and `Ruler.matches` reject each trailer shape; the second-value case carries the compiler's message; whitespace-only trailers still compile and match. Against unmodified `main` (b877fa9) the four rejection tests fail with `check must reject: {"orderId":[{"exists":true}]}}` / `matchesRule must reject: ...` and the whitespace tests pass. Commit 2: `MachineTest.testTrailingContentIgnoredOnlyWhenConfigured` — the default machine rejects every trailer through all four `addRule` input types and stays empty; the configured machine adds, matches, and deletes the same rules; `check()` rejects regardless. With the option wired but the compilers ignoring it, that test fails on the first lenient `addRule`. `mvn verify` on this branch: 804 tests green, checkstyle included.

**Compatibility.** This narrows the accepted input. A rule that was stored with trailing content compiles today and fails to compile after this change unless the machine is built with the option above, so a consumer that rebuilds machines from stored rules should scan the store with the new `check()` before upgrading, and use the option only for the window in which it does. Release notes should call this out; I lean toward a minor bump (2.2.0) for that reason.

### DIFF
--- a/README.md
+++ b/README.md
@@ -653,6 +653,26 @@ Machine machine = Machine.builder()
     .build();
 ```
 
+#### withTrailingContentIgnored
+Default: false (since 2.2.0)
+
+A rule is one JSON object. Since 2.2.0, `addRule()` and `deleteRule()` reject a rule
+followed by anything other than whitespace — an extra `}` or `]`, a `,`, a bare word, a
+second object — with the parse error, and `RuleCompiler.check()` / `JsonRuleCompiler.check()`
+report it. Releases before 2.2.0 stopped reading at the rule's closing brace and ignored
+whatever came after it.
+
+Set this to true only for a machine rebuilt from rules that were stored while the old
+reading was in force and have not been re-validated yet; it restores that reading for
+`addRule()` and `deleteRule()` on that machine. `check()` is not affected — it always
+rejects trailing content, so it is the right tool for finding such rules in a store:
+
+```java
+Machine machine = Machine.builder()
+    .withTrailingContentIgnored(true)
+    .build();
+```
+
 ### addRule()
 
 All forms of this method have the same first argument, a String which provides
--- a/src/main/software/amazon/event/ruler/GenericMachine.java
+++ b/src/main/software/amazon/event/ruler/GenericMachine.java
@@ -451,9 +451,9 @@ public class GenericMachine<T> {
      */
     public void addRule(final T name, final String json) throws IOException {
         try {
-            JsonRuleCompiler.compile(json, configuration.isRuleOverriding()).forEach(rule -> addPatternRule(name, rule));
+            JsonRuleCompiler.compile(json, configuration.isRuleOverriding(), configuration.isIgnoreTrailingContent()).forEach(rule -> addPatternRule(name, rule));
         } catch (JsonParseException e) {
-            addPatternRule(name, RuleCompiler.compile(json, configuration.isRuleOverriding()));
+            addPatternRule(name, RuleCompiler.compile(json, configuration.isRuleOverriding(), configuration.isIgnoreTrailingContent()));
         }
     }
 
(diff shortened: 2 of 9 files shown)


timbray:
Yes, a version bump is definitely called for.

fym-rgb:
Will add a lenient mode override as an option, to keep some backward compatibility.

leezachydev (approved):

PR #280, opened 2026-09-17
fix: Return every NameState an anything-but pattern reaches

### LINKED ISSUE #259: Machine.deleteRule ghost/strand defect also affects anything-but wildcard patterns (follow-up to #255/#256)
## Describe the bug

#256 fixed `deleteRule` isolation for rules sharing byte-identical wildcard clauses (ghost: deleted rule keeps matching; strand: surviving sibling stops matching — see #255). The same defect class also affects `anything-but: {wildcard: ...}` patterns and is not covered by that fix.

Mechanism: `ByteMachine.findAllPatterns` routes `ANYTHING_BUT_WILDCARD` (and the other `ANYTHING_BUT*` types) to the `default:` single-NameState path. Anything-but-wildcard values are wildcard-parsed into the ByteMachine, so the add path's refusal to reuse indeterminate-prefix / self-loop composite states (`canReuseNextByteState`) applies — byte-identical anything-but-wildcard clauses accumulate one `NameState` per rule over shared transitions, and the single-NameState delete path picks an arbitrary one (ghost) or tears down shared transitions still referenced by a sibling (strand).

This is pre-existing: the repro below fails identically on `main` before and after #256 — #256 did not introduce or worsen it.

## To Reproduce

Non-deterministic ~50% per trial (HashSet iteration order), so run over many fresh machines:

```java
@Test
public void anythingButWildcard_sharedClause_deleteIsolation() throws Exception {
    String ruleJson = "{\"name\":[\"test\"],\"properties.foo\":[{\"anything-but\":{\"wildcard\":\"*bar*\"}}]}";
    String event = "{\"name\":\"test\",\"properties\":{\"foo\":\"nomatchhere\"}}";

    for (int i = 0; i < 200; i++) {
        Machine machine = Machine.builder().build();
        machine.addRule("rule1", ruleJson);
        machine.addRule("rule2", ruleJson);
        assertEquals(2, machine.rulesForJSONEvent(event).size());

        machine.deleteRule("rule1", ruleJson);

        List<String> after = machine.rulesForJSONEvent(event);
        assertFalse("rule1 should be deleted (ghost)", after.contains("rule1"));
        assertTrue("rule2 should still match (strand)", after.contains("rule2"));
    }
}
```

Also reproduces with different-shaped rules sharing the same anything-but-wildcard clause on one field.

## Expected behavior

Same as #255: `deleteRule` removes only the named rule; siblings sharing the clause keep matching; result independent of add order.

## Proposed fix

Extend `findAllPatterns` to cover the `ANYTHING_BUT*` types (and audit `NUMERIC_RANGE`). The anything-but find paths are multi-value — each value in the set is looked up individually and the results are asserted to converge to a single NameState (`findAnythingButPattern` / `findAnythingButValuesSetPattern`) — so the extension needs per-value `findAllMatchPattern` with union semantics rather than the simple single-value delegation #256 added for value patterns. The teardown guard from #256 (`noNameStateContainsPattern`) already generalizes.

I'll take this as a stacked change on top of #256.

## Environment

- event-ruler current `main` (also reproduces on the #256 branch and on 1.8.x)

leezachydev:
### Issue #, if available:

Fixes #259

### Description of changes:

When rules share an anything-but pattern on a key, the shared byte transitions can lead to more than one `NameState`, one branch per sub-rule. `deleteRule` cleans a pattern up by walking the `NameState`s that `ByteMachine.findAllPatterns` reports, so it must report all of them. `findAnythingButPattern` and `findAnythingButValuesSetPattern` instead collapsed the discovered branches to a single `NameState` (asserting `size() == 1`), and `findAllPatterns` returned only that one via `asSet(...)`. `GenericMachine.deleteStep` then never visited every branch: it could skip the branch belonging to the rule being deleted, leaving that rule's own sub-rule entry stranded so it kept matching (a ghost), or it could tear down a shared transition and take the surviving rule's branch with it (over-delete). Either way the machine was never fully torn down (`isEmpty()` stayed false).

The `size() == 1` assert happened to hold for the value, prefix, suffix, and ignore-case anything-but types, which reach a single shared `NameState`. `ANYTHING_BUT_WILDCARD` reaches several, and it is the one that broke in practice.

PR #256 taught `findAllMatchPattern` to return every `NameState` a value pattern reaches, but that union fix never extended to the anything-but paths. This change closes that gap:

- `findAnythingButPattern` and `findAnythingButValuesSetPattern` now aggregate `findAllMatchPattern` over each excluded value and return the full `Set<NameState>`, dropping the `size() == 1` collapse.
- `findAllPatterns` returns those sets directly instead of wrapping a single `NameState` in `asSet(...)`.
- `findMatchPattern`, whose only callers were those two helpers, is removed.

Matching is unchanged: the additional `NameState`s were always reachable during the forward match walk; they were only hidden from the delete-side reverse lookup.

**Tests.** `AnythingButSharedDeleteTest` adds three cases. `WHEN_SameKeyRulesShareAnythingButWildcard` adds two rules sharing an anything-but-wildcard on one key, deletes one and asserts the survivor still matches its allowed value and still excludes the wildcard value, then deletes both and asserts `isEmpty()`. `WHEN_DivergingRulesShareAnythingButWildcard` gives the two rules the same wildcard on key `a` but different `b` values, so the shared `a` transition reaches distinct `NameState`s that continue to the `b` patterns; deleting one must not tear down the other's branch. `WHEN_RulesShareAnythingButPrefix` is the same shape on a prefix and guards the anything-but paths that already collapsed to a single shared `NameState`. Against unmodified `main` (`6efe434`) the two wildcard tests fail and the prefix test passes; with the fix, `mvn test` reports `Tests run: 834, Failures: 0, Errors: 0, Skipped: 0` and BUILD SUCCESS.

This is a teardown-only correctness fix. No public API change and no behavior change to matching.

#### Dependencies (if the change adds one, upgrades one, or changes the scope of one):

None. No dependency added, upgraded, or scope-changed.

```
No hot-path change: this only corrects delete-side NameState discovery, not the forward match walk. Benchmark output to be fetched from the Java build check (Run benchmarks) on this PR.
```


fym-rgb:
The implementation looks correct.

Before approval:

1. Please add to the PR: “By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.”

2. In the [diverging-rule test comment](https://github.com/aws/event-ruler/blob/9e118f8638a1a92d0d1e490c964e90259b4384d4/src/test/software/amazon/event/ruler/AnythingButSharedDeleteTest.java#L58-L66), the states after `a` are not terminal; both rules still require `b`. “The shared `a` transition reaches distinct NameStates that continue to the `b` patterns” would be accurate. Please remove the terminal-only generalization from the class comment and description too.

3. Please reconcile the description's 838-test count with the exact command and output. My clean run found 834. The failure explanation also switches from “the other's” entry to “the deleted rule”: in the continued-matching case, the skipped branch belongs to the rule being deleted.

4. Optional test improvement: after deleting `r1`, assert that `{"a":"foo"}` still matches no rule. That checks the surviving rule's exclusion as well as its allowed `bar` case. My additional checks passed this case; it is coverage hardening, not a defect I found.

---
###### GenAI-assisted Review

fym-rgb (approved):
lgtm

